PR #27, opened 2023-06-17
Clean up/re-organize note types and add some new things to them

TheNathannator:
Reorganized a number of things to make them a little cleaner, added new note values, types, and flags, added Pro Guitar notes, and fixed the child notes list never being initialized.

Only thing I have specific comments on is regarding the removal of the `previousNote` parameter from the note constructors, the chart loading infrastructure I'm working on doesn't use or need it and the field it assigns to is public anyways lol

### DIFF
--- a/YARG.Core/Chart/Events/ChartEvent.cs
+++ b/YARG.Core/Chart/Events/ChartEvent.cs
@@ -5,13 +5,13 @@ namespace YARG.Core.Chart
     /// </summary>
     public abstract class ChartEvent
     {
-        public double Time { get; }
+        public double Time       { get; }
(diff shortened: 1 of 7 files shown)


RileyTheFox (changes requested):
Include IsStrum, IsTap, IsHopo properties for the GuitarNote and ProGuitarNote. This just makes it easier to use than `if(note.NoteType == Strum)` all the time, versus just doing `if(note.IsStrum)`
The types should also be settable in some way to account for future note modifiers (all strums, all taps etc)

RileyTheFox (approved):

PR #29, opened 2023-07-02
Implement loading/conversion of MoonSongs to SongCharts

TheNathannator:
Big monumental step towards the new engine!

The song loading has been abstracted in a way that should make it easier to switch to another parser or support multiple parsers at the same time. The only references to any Moonscraper types now, outside of within Moonscraper itself, are in the MoonSong loader and in the parser unit tests.

A good number of changes to the chart storage types have been made along the way:

- `InstrumentTrack` no longer contains a `MaxDifficulty` field, though this could be re-added later if it would be convenient enough. Removed it because it seems less convenient/necessary than it did when I first added it.
- `SongChart` is now fully default-initialized and can be created without going through a chart file.
- Vocal tracks now have their own dedicated storage types, since they work very differently from the other tracks lol
- The 6-fret instrument/game mode definitions were uncommented, not much additional work was needed to load them in.
- `Note` has been turned into a recursive generic type (`Note<TNote> where TNote : Note<TNote>`), to remove the need to type-test on the previous/next notes or child note list to get the true type. Any methods that need to work with just `Note`s can be written as generic methods in the same fashion.
- `VocalNote`'s pitch-over-time storage has been reworked to use the child note list, as this greatly simplifies vocals parsing (which is already complex enough lol).
- Some various other things were reworked in `VocalNote`, and doc comments were added since some things may not be clear at a first glance.

And to top things off, I fixed a Moonscraper parsing bug I found when testing vocals loading.

### DIFF
--- a/YARG.Core/Chart/Loaders/ISongLoader.cs
+++ b/YARG.Core/Chart/Loaders/ISongLoader.cs
@@ -14,6 +14,6 @@ namespace YARG.Core.Chart
         InstrumentTrack<GuitarNote> LoadGuitarTrack(Instrument instrument);
         InstrumentTrack<ProGuitarNote> LoadProGuitarTrack(Instrument instrument);
         InstrumentTrack<DrumNote> LoadDrumsTrack(Instrument instrument);
-        InstrumentTrack<VocalNote> LoadVocalsTrack(Instrument instrument);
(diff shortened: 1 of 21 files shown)


RileyTheFox (commented):
very good

PR #32, opened 2023-08-07
WIP: Song scanning

sonicfind:


YARG.Core/InstrumentEnums.cs
```diff
@@ -58,6 +58,18 @@ public enum Instrument
...
+    /// <summary>
+    /// Available Drum types. Unknown values for preparse purposes
+    /// </summary>
+    public enum DrumType
+    {
+        FOUR_LANE,
+        FOUR_PRO,
+        FIVE_LANE,
+        UNKNOWN,
+        UNKNOWN_PRO,
```
TheNathannator, line 70: Preferred style for enums is PascalCase, not CONST_CASE.

YARG.Core/Song/Deserialization/Ini/IniModifier.cs
```diff
@@ -0,0 +1,308 @@
...
+            union.d = value;
+        }
+        public IniModifier(ulong dub1, ulong dub2)
+        {
+            type = ModifierType.UINT64ARRAY;
+            union.ulArr[0] = dub1;
+            union.ulArr[1] = dub2;
+        }
+
+        public SortString SORTSTR
```
TheNathannator, line 115: Property names should be PascalCase as well, preferably also without shortening.
```diff
- public SortString SORTSTR
+ public SortString SortString
```
This one's a little odd since it results in the same name as the type lol, but that happens sometimes.

YARG.Core/Song/Deserialization/Ini/IniHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+                { "diff_keys",                            new("diff_keys", ModifierNodeType.INT32) },
+                { "diff_keys_real",                       new("diff_keys_real", ModifierNodeType.INT32) },
+                { "diff_keys_real_ps",                    new("diff_keys_real_ps", ModifierNodeType.INT32) },
+                { "diff_rhythm",                          new("diff_rhythm", ModifierNodeType.INT32) },
+                { "diff_rhythm_ghl",                      new("diff_rhythm_ghl", ModifierNodeType.INT32) },
+                { "diff_vocals",                          new("diff_vocals", ModifierNodeType.INT32) },
+                { "diff_vocals_harm",                     new("diff_vocals_harm", ModifierNodeType.INT32) },
+                { "drum_fallback_blue",                   new("drum_fallback_blue", ModifierNodeType.BOOL) },
+
+                { "early_hit_window_size",                new("early_hit_window_size", ModifierNodeType.STRING) },
```
TheNathannator, line 71: Some of these .ini tags are an odd inclusion, do we really need to handle *all* of these? No point checking what we never use.

YARG.Core/Song/Deserialization/Ini/IniHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+                { "song_length",                          new("song_length", ModifierNodeType.UINT64) },
+                { "star_power_note",                      new("multiplier_note", ModifierNodeType.INT32) },
+                { "sub_genre",                            new("sub_genre", ModifierNodeType.SORTSTRING) },
+                { "sub_playlist",                         new("sub_playlist", ModifierNodeType.SORTSTRING) },
+                { "sustain_cutoff_threshold",             new("sustain_cutoff_threshold", ModifierNodeType.INT64) },
+                { "sysex_high_hat_ctrl",                  new("sysex_high_hat_ctrl", ModifierNodeType.BOOL) },
+                { "sysex_open_bass",                      new("sysex_open_bass", ModifierNodeType.BOOL) },
+                { "sysex_pro_slide",                      new("sysex_pro_slide", ModifierNodeType.BOOL) },
+                { "sysex_rimshot",                        new("sysex_rimshot", ModifierNodeType.BOOL) },
+                { "sysex_slider",                         new("sysex_slider", ModifierNodeType.BOOL) },
```
TheNathannator, line 129: Same sentiment with these tags. There's no reason to be able to enable or disable SysEx events, we're never gonna use these lol

YARG.Core/Song/Deserialization/YARGBinaryReader.cs
```diff
@@ -0,0 +1,397 @@
...
+
+namespace YARG.Core.Song.Deserialization
+{
+    public enum Endianness
+    {
+        LittleEndian = 0,
+        BigEndian = 1,
+    };
+
+#nullable enable
```
TheNathannator, line 17: Probably something we should enable for the whole project if we're gonna use it lol

YARG.Core/Song/Deserialization/YARGFile.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace YARG.Core.Song.Deserialization
+{
+    public unsafe class YARGFile
+    {
+        private readonly GCHandle handle;
+        private readonly byte* _data;
+        private readonly int _length;
+
+        public byte* Data => _data;
```
TheNathannator, line 13: Really should just use a `byte[]` or `ReadOnlyMemory<byte>` here or something, GCHandles are expensive and you can easily get a `byte*` from a `byte[]` by using a `fixed` statement where they're needed.

TheNathannator (commented):
Just a few quick comments I have from taking a quick look, haven't gotten a chance to review thoroughly lol

YARG.Core/Song/Deserialization/YARGBinaryReader.cs
sonicfind: **For the love of God, please figure out how to do that**

To be real, yes, it does. Turning it on actually led to me finding a potential bug before production.

TheNathannator (changes requested):
Took a much more thorough look this time, I definitely think a lot of this is over-engineered but it's nothing I wouldn't be able to work with lol

A lot of these requests are just making constants for things and de-duplicating code where possible. Fast code is nice and good, but not at the expense of code readability. I don't care if we lose 5% performance by making things a lot more readable.

I know this is still a work-in-progress, but I wanted to get these comments out while I have the time.

YARG.Core/Song/Metadata/SongMetadata.SongPackedRBCON.cs
```diff
@@ -0,0 +1,319 @@
...
+                string midiDirectory = conFile[midiListing.pathIndex].Filename;
+
+                moggListing = conFile[location + ".mogg"];
+
+                if (!location.StartsWith($"songs/{nodeName}"))
+                    nodeName = midiDirectory.Split('/')[1];
+
+                string genPAth = $"songs/{nodeName}/gen/{nodeName}";
+                miloListing = conFile[genPAth + ".milo_xbox"];
+                imgListing = conFile[genPAth + "_keep.png_xbox"];
```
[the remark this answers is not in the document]
sonicfind: The index will return a null if the listing doesn't exist at that spot. I'll replace the method without something more descriptive

YARG.Core/Song/Preparsers/Midi/MidiDrumPreparser.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        protected override bool ProcessSpecialNote_Off()
+        {
+            if (note.value != 95)
+                return false;
+
+            if (notes[3, 1])
+                validations |= 24;
+            return true;
```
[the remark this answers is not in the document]
sonicfind: Yea, don't worry. The midi preparsers are what I'm planning to go back to once I got everything else in an operational state. It's just that since there's a deadline, there's only so much I could do right away.

YARG.Core/Song/Deserialization/YARGBinaryReader.cs
TheNathannator: It's just a tag in the project file, super easy to enable:
```xml
<PropertyGroup>
  ...
  <Nullable>true</Nullable>
  ...
</PropertyGroup>
```

We'll just need to add all of the annotations where needed when we enable it, would be best done as its own set of changes.

YARG.Core/Song/Deserialization/YARGFile.cs
TheNathannator: > you need to use a pointer and explicit length for nested regions

This is precisely what `Span<T>` and `Memory<T>` are for, you can easily get one from a byte array by using the `AsSpan()`/`AsMemory()` extension methods (or just assigning it directly to a variable of that type), and you can slice them up using the range operator or the `Slice()` method.

sonicfind: Except Memory<T> does not have an indexer and must converted to a span on every call you try to use it on. That is *not* a free operation - much like `fixed`.
And I wish I could use a Span<T> as a member variable, but that requires that the readers be `ref struct` which brings a whole other set of issues. SpanSplitter works as it does because every use-case initializes a local variable specifically.

YARG.Core/Song/Metadata/PartValues.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+
+namespace YARG.Core.Song
+{
+    [Serializable]
+    public struct PartValues
+    {
+        public byte subTracks;
+        public sbyte intensity;
```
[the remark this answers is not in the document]
sonicfind: It just ensures 1. a sensible range, and 2. that later programmers and/or artists down the road don't have to worry about the prospect of someone trying to make a song with like "hurr durr 2,147,486,647" as the intensity.

0 - 127 as an easy to understand baseline (more for ini songs); full int range for CONs (used in playlist ordering) because they already go higher.

YARG.Core/Song/Deserialization/YARGFile.cs
sonicfind: That being said, I'll go ahead do Memory<T> for right now

YARG.Core/Song/Preparsers/Midi/MidiDrumPreparser.cs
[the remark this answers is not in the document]
sonicfind: Did a bunch of constants at least

YARG.Core/Song/Deserialization/YARGFile.cs
TheNathannator: > and must be converted to a span on every call you try to use it on. That is *not* a free operation - much like `fixed`.

Not as far as I can tell from reading the [code](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Memory.cs#L307); at least in modern .NET, it's just getting the required reference from the array directly using `MemoryMarshal` (which reaches into [internal stuff](https://github.com/dotnet/runtime/blob/main/src/mono/System.Private.CoreLib/src/System/Runtime/InteropServices/MemoryMarshal.Mono.cs#L23) to do its job), and doesn't do any pinning.

After some investigation, it also seems `fixed` doesn't do the same pinning that `GCHandle` does either, the JIT and GC work together to be more efficient about things. [1](https://stackoverflow.com/a/22204244) [2](https://github.com/ralfbiedert/interoptopus/issues/16)

As a final note, `Memory<T>` tracks its data by using an opaque `object` field (no boxing occurs since all types assigned to it are already reference types), alongside index and length fields for where in the data it refers to, there is no copying that happens when you `Slice()`. [1](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Memory.cs#L29-L31)

~~spent far too long researching this 'cause i was curious myself lol~~

YARG.Core/Song/Deserialization/YARGChartFileReader.cs
```diff
@@ -0,0 +1,306 @@
...
+            new(Encoding.ASCII.GetBytes("DoubleGuitar]"), NoteTracks_Chart.DoubleGuitar ),
+            new(Encoding.ASCII.GetBytes("DoubleBass]"),   NoteTracks_Chart.DoubleBass ),
+            new(Encoding.ASCII.GetBytes("DoubleRhythm]"), NoteTracks_Chart.DoubleRhythm ),
+            new(Encoding.ASCII.GetBytes("Drums]"),        NoteTracks_Chart.Drums ),
+            new(Encoding.ASCII.GetBytes("Keyboard]"),     NoteTracks_Chart.Keys ),
+            new(Encoding.ASCII.GetBytes("GHLGuitar]"),    NoteTracks_Chart.GHLGuitar ),
+            new(Encoding.ASCII.GetBytes("GHLBass]"),      NoteTracks_Chart.GHLBass ),
+            new(Encoding.ASCII.GetBytes("GHLRhythm]"),    NoteTracks_Chart.GHLGuitar ),
+            new(Encoding.ASCII.GetBytes("GHLCoop]"),      NoteTracks_Chart.GHLBass ),
+        };
```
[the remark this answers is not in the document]
sonicfind: While I understand, we should really be actively discouraging the usage of anything other that UTF-8 (although we should handle the Extended ASCII case). Due to how .chart data is formatted, anything other than UTF-8 is just massively bloating filesize. I'd rather have a check on the scanning end that throws an exception (not error) if a UTF-16/32 or other BOM is recognized so that authors don't inadvertently use it.

YARG.Core/Song/Deserialization/YARGChartFileReader.cs
```diff
@@ -0,0 +1,306 @@
...
+            {
+                if (descriptor.Length != length) return false;
+                for (int i = 0; i < length; ++i)
+                    if (descriptor[i] != data[start + i]) return false;
+                return true;
+            }
+
+            long position = reader.ReadInt64();
+            if (position < tickPosition)
+                throw new Exception($".Cht/.Chart position out of order (previous: {tickPosition})");
```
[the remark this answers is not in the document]
sonicfind: ah shit... DON'T LOOK

YARG.Core/Song/Deserialization/YARGDTAReader.cs
```diff
@@ -0,0 +1,226 @@
...
+                        throw new Exception("Text error - no quotes allowed");
+                }
+                else if (ch == '\'')
+                {
+                    if (inApostrophes)
+                        break;
+                    if (!inSquirley && !inQuotes)
+                        throw new Exception("Text error - no apostrophes allowed");
+                }
+                else if (ch <= 32)
```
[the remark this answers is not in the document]
sonicfind: The multiple-multiple levels of indirection on top of a bunch char-type only redundancy checks that char.IsWhitespace brings has always been the sight of a bunch of performance headaches (and how it has to cast to char every time). Like, more than 20%. Not even C++ fixes that shit either.

YARG.Core/Song/Deserialization/YARGBinaryReader.cs
sonicfind: I'd rather have this handled as more of a final step for YARG.Core in general, tbh. It's more of a general scope.

TheNathannator: That's what I was saying lol, this was just a general comment.

YARG.Core/Song/Deserialization/YARGDTAReader.cs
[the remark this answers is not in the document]
TheNathannator: interesting, yeah nvm then lol

```
|       Method | character |      Mean |     Error |    StdDev |    Median |
|------------- |---------- |----------:|----------:|----------:|----------:|
| IsWhiteSpace | '\t'      | 0.3537 ns | 0.0094 ns | 0.0088 ns | 0.3570 ns |
|   Is32OrLess | '\t'      | 0.0037 ns | 0.0049 ns | 0.0046 ns | 0.0013 ns |
| IsWhiteSpace | ' '       | 0.3528 ns | 0.0069 ns | 0.0064 ns | 0.3526 ns |
|   Is32OrLess | ' '       | 0.0075 ns | 0.0068 ns | 0.0064 ns | 0.0060 ns |
```

PR #33, opened 2023-08-10
Increase Starpower with Whammy

JayDiddyThaGOAT:
Noticed that whammy doesn't increase star power during a star power phrase. This is the fix for it.


RileyTheFox (changes requested):
This would add star power for each sustain that is active. Meaning if 3 sustains were active, you would get 3x the star power, which is not how that works.

Move it outside the for loop and it should be fine

RileyTheFox:
This was also intentionally not added as whammy has not been added in YARG's side completely. Sustains are just kind of placeholder right now

PR #51, opened 2023-08-14
Scan Logging

sonicfind:


YARG.Core/Song/Cache/CacheHandler.cs
```diff
@@ -53,6 +53,7 @@ public SongCache RunScan(bool fast)
         public const int CACHE_VERSION = 23_08_12_01;
 
         public readonly List<object> errorList = new();
+        public readonly List<string> logs = new();
```
TheNathannator, line 56: Any reason you're collecting log messages like this instead of using `YargTrace`'s debug logging methods directly?

This gets rid of the relationship between the message and when it actually happened, and if any errors get logged before these messages do, it will be impossible to correlate the error and this logging together.

`YargTrace` also has `Debug` variants of the logging methods that are only included in debug builds, which verbose/debug logging like this should preferably use (at least for now).

YARG.Core/Song/Cache/CacheHandler.cs
```diff
@@ -238,6 +240,9 @@ private void WriteBadSongs()
                     case ScanResult.PossibleCorruption:
                         writer.WriteLine("Possible corruption of a queried midi file");
                         break;
+                    case ScanResult.PathTooLong:
+                        writer.WriteLine("Path too long the Windows Filesystem (path limitation can be changed in registry settings)");
```
TheNathannator, line 244: Missed a word here lol
```diff
- Path too long the Windows Filesystem (path limitation can be changed in registry settings)
+ Path too long for the Windows Filesystem (path limitation can be changed in registry settings)
                ^
```

TheNathannator (changes requested):

YARG.Core/Song/Cache/CacheHandler.cs
sonicfind: fuck

YARG.Core/Song/Cache/CacheHandler.cs
sonicfind: Does YARGTrace handle multithreading by itself?

TheNathannator: YargTrace itself doesn't, but both console output and Unity's debug logging are thread safe. The only thing that might need to be synchronized is the list of listeners, but in practice that doesn't really matter since only one listener is ever added lol

TheNathannator (approved):
Looks good now!

PR #61, opened 2023-08-17
Fix mogg audio values index out of bounds errors with song_updates

sonicfind:


TheNathannator:
lol forgot about this, sorry about that!

PR #74, opened 2023-09-18
Some Reader Touchups

sonicfind:


YARG.Core/Song/Deserialization/TXTReader/ITXTReader.cs
```diff
@@ -7,23 +7,21 @@ namespace YARG.Core.Song.Deserialization
     public interface ITXTReader
     {
         public static readonly Encoding Latin1 = Encoding.GetEncoding(28591);
-        public bool ReadBoolean(ref bool value);
```
TheNathannator, line 10: As a quick note, I personally think `ReadBoolean(out bool)` should have stayed, it's not really good design to hide away invalid data that might be encountered. Though invalid data wasn't really checked for properly anyways lol

PR #75, opened 2023-09-18
Enable nullable analysis and fix all nullable warnings

TheNathannator:
There might be a couple things I missed, but this should be pretty much everything according to the absence of warnings.

The [PolySharp](https://github.com/Sergio0694/PolySharp) package was brought in to polyfill some nullable analysis attributes that are missing from .NET Standard 2.1. This package also allows us to use C# 9 language features that rely on other missing attributes, such as init-only setters.


TheNathannator:
Leaving this overnight for review, feel free to merge ahead of me if it's good and nothing's missing.

PR #94, opened 2023-10-11
Split YARGMidiReader Into Separate YARGMidiFile & YARGMidiTrack classes

sonicfind:


TheNathannator:
Concerned this'll conflict too much with my own MIDI reader stuff I'm working on, as now these files are in the exact same spot as my own. Not gonna be an easy rebase lol

TheNathannator:
Decided not to continue with my parser. Some good ideas, but it's too tailored towards a specific use case, and the cons of that outweigh the pros.

PR #104, opened 2023-10-30
Improve Drum Preparse Handling

sonicfind:
+ Fixes drum scans for CON update files being overridden by the original midi file.

+ Reorder and restructure songini preparse functions

+ Fix ParseSettings not receiving the drumType

### DIFF
--- a/YARG.Core/Song/Cache/CacheHandler.cs
+++ b/YARG.Core/Song/Cache/CacheHandler.cs
@@ -48,7 +48,7 @@ namespace YARG.Core.Song.Cache
         /// Format is YY_MM_DD_RR: Y = year, M = month, D = day, R = revision (reset across dates, only increment
         /// if multiple cache version changes happen in a single day).
         /// </summary>
-        public const int CACHE_VERSION = 23_10_30_01;
+        public const int CACHE_VERSION = 23_10_30_02;
 
         private static readonly object dirLock = new();
         private static readonly object fileLock = new();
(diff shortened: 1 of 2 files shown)


jgoyvaerts:
Shouldn't we write unit tests for this stuff?

sonicfind:
Unit testing an entire complex system that is extremely dependent on slight differences  in user provided data (a lot of which are very minuscule) is a fools errand. The best test you can ever do is with real samples of data, of which you know the desired result.

It's why SongScanning test exists really. You can provide the folder/song to scan and compared the resulting entry or entries to what you expected to receive.

jgoyvaerts:
I guess unit testing was a poor choice of word. I mean an integration test which loads a (custom crafted) midi file and verifies the cymbals are correct.

The complexity of the system should not be an excuse not to write any kind of tests, on the contrary. The more complex the system, the more useful tests can be.

sonicfind:
Integration makes more sense, yea. I guess the issue with how it works is that you can't just have *one* file. You'd really need dozens or hundreds of fully arranged directories to contain all the possible test cases that you'd need to consider. Some with ini files, some with no audio, some with inis that *fail* to parse, some with ini data that contradicts the midi files, midi files with bad data, midi files with no data, .chart files and all those bring, .chart files *in the same folder as .mid files*, CON files with their own quirks.

Trying to set that up takes an enormous amount of time, of which for a lively-updated game, and for a feature of which had a short time limit on pushing out initially to start with, we don't really have. Furthermore, we'd probably need to have a completely separate integration testing repo with all those mock files so as to not force every person to download the whole suite of files that they probably won't ever even touch.

TheNathannator:
Unit/integration testing is something we should definitely set up more often, the main reason we haven't is just because it takes a lot of time and effort which could also be spent elsewhere lol, especially on edge cases.

You're always welcome to PR tests if you get around to them, I'm personally not very experienced with them so I don't think about it often.

PR #105, opened 2023-10-31
Decrease Number of Filestream `Open` Calls Made to CONs During Scanning

sonicfind:
Using a lock on a shared filestream can save 100s of calls to the OS to open CON files during a fresh scanning process. Combines with the benefit of not having lingering listings hogging memory after the scan is complete.

In other words, reintroduce a CONFile class type - but without the baggage. Yippee.

HDDs may see more benefit from this, but I can't check that myself.


YARG.Core/IO/ConHandler/CONFile.cs
```diff
@@ -5,8 +5,36 @@
...
+                    return listing;
+            }
+            return null;
+        }
+
+        public void Dispose()
+        {
+            Listings.Clear();
+            Stream.Dispose();
+        }
```
TheNathannator, line 36: Was gonna knock this for not being a full `IDisposable` pattern lol, but so long as only managed resources need to be cleaned up, this is fine.

sonicfind: Actually, I could just *not* use IDisposable

TheNathannator: Well, it'd still be good to have to show intent (even though it still has the Dispose method). Though it might not use unmanaged resources directly, `FileStream` does. `IDisposable` is the standard method of signalling this, and it's generally what you should follow.

This is more in the realm of API design though, which I do think should be considered just a little bit more in the scanning code to prevent certain issues, but it's ultimately not a high priority.

sonicfind: fair enough

sonicfind: Considering some plans I have that *will* enforce more of the IDispoable interface in scanning code, I'm down either way

PR #107, opened 2023-11-02
Make MIDI readers choose/auto-detect correct text encoding for lyrics/text events

raphaelgoulart:
Previously, the MIDI readers would assume lyrics/text events are UTF-8 encoding by default. However, that is not true to all MIDIs. The proposed solution goes as thus:

- RBCONs are set to Latin-1 encoding. The `encoding` parameter on the DTA applies only to the DTA itself; RB MIDIs are always Latin-1 encoded.
- .ini MIDIs are read with strict (throws error on invalid character) UTF-8 parsing rules.
  - If that fails, the MIDI is re-read with Latin-1 encoding.
- .chart reading is untouched.

This shouldn't affect performance in any significant way; scanning won't be affected, .charts and RBCONs won't be affected, and most MIDIs will be valid UTF-8 (intentionally or not; english-only Latin-1 is also perfectly valid UTF-8). The only case in which the MIDI would be read twice is old/C3CON Tools converted non-english songs, but the benefit of their lyrics being read correctly more than makes up for the negligible performance loss of reading the MIDI twice (since song loading is near-instant anyway).

### DIFF
--- a/YARG.Core/Chart/MidiSettings.cs
+++ b/YARG.Core/Chart/MidiSettings.cs
@@ -27,16 +27,4 @@ namespace YARG.Core.Chart
             TextEncoding = YARGTextContainer.Latin1,
         };
 	}
-
-    public static class MidiSettingsUTF8
-    {
-        public static readonly ReadingSettings Instance = new()
-        {
-            InvalidChunkSizePolicy = InvalidChunkSizePolicy.Ignore,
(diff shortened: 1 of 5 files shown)


sonicfind (approved):

TheNathannator:
Only comment I have is the Latin-1 settings don't strictly need a separate container class, could rename `Instance` to `UTF8` and merge the Latin1 version in as a `Latin1` field.

Otherwise, looks good!

PR #108, opened 2023-11-03
Auto-generate venue lighting incase it doesn't exist

raphaelgoulart:
This code reads a JSON file containing info on which lighting presets/post-procs to use for which sections, and generates a lighting track based on that info, if the song currently being loaded doesn't have it.
Implementation loosely based on MagmaC3's, detailed in http://docs.c3universe.com/rbndocs/index.php?title=Venue_Autogeneration
This needs implementation on the YARG side as well, but the game should still behave the same without it (i.e. songs loading normally, without venue autogeneration).

Example auto-generation preset JSON file, for testing purposes: [DefaultPreset.json](https://github.com/YARC-Official/YARG.Core/files/13253806/DefaultPreset.json)
Demo video: https://www.youtube.com/watch?v=jd588FNhSOE (for any `[next]` counter that resets to 0 for no reason -- that's the blend-in lighting event, not a `[first]` keyframe)

Suggestion for implementation on the YARG side:
- Have built-in presets in some game folder, like `StreamingAssets/autogen` or similar;
- Read custom presets from `PersistentDataPath/autogen` or similar;
- List them in a dropdown in the settings, either near to other venue-related settings, or on the Presets tab;
- Pass the selected preset as a parameter for the Song.LoadChart() function.

Please let me know if any changes are needed.

### DIFF
--- a/YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Venue.cs
+++ b/YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Venue.cs
@@ -26,7 +26,7 @@ namespace YARG.Core.Chart
             { VENUE_PERFORMER_KEYS,   Performer.Keyboard },
         };
 
-        private static readonly Dictionary<string, LightingType> LightingLookup = new()
+        public static readonly Dictionary<string, LightingType> LightingLookup = new()
         {
             // Keyframed
             { VENUE_LIGHTING_DEFAULT,     LightingType.Default },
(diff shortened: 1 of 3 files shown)


RileyTheFox:
I think all venue generation should be done on YARG's side, as YARG.Core doesn't care about visual information. The only venue things in YARG.Core are for reading the data in a chart file. This also removes the need for additional parameters on chart parsing functions (`LoadChart` for example) which I don't like as this info is irrelevant for the purposes YARG.Core serves.

Also small code review but a few parts of the code are using the wrong curly brace formatting (opening should be placed on a new line) and there are a lot of places where `var` should be used instead of the full type name.

PR #109, opened 2023-11-04
Refactor Venue lookup dictionaries to public class

raphaelgoulart:
Prerequisite to clean up https://github.com/YARC-Official/YARG/pull/588


TheNathannator (approved):
Should do fine for now.

I've been thinking of restructuring some of these definitions regardless, but that won't happen until I get around to writing some custom chart parsers, which probably isn't necessary after sonicfind's optimizations of the Moonscraper code lol

PR #112, opened 2023-11-09
Some YARGMidiTrack Encapsulation Changes

sonicfind:
The removal of the YARGBinaryReader variable and instead having YARGMidiTrack handle its own data is the biggest change

### DIFF
--- a/YARG.Core/IO/Midi/YARGMidiTrack.cs
+++ b/YARG.Core/IO/Midi/YARGMidiTrack.cs
@@ -79,6 +79,9 @@ namespace YARG.Core.IO
             }
         }
 
+        private const int CHANNEL_MASK = 0x0F;
+        private const int EVENTTYPE_MASK = 0xF0;
+
         public bool ParseEvent(bool parseVLQ)
         {
             _trackPos += _event.Length;
@@ -101,8 +104,8 @@ namespace YARG.Core.IO
                 _trackPos++;
                 if (type < MidiEventType.SysEx)
                 {
-                    _event.Channel = _running.Channel = (byte) (tmp & 15);
-                    _event.Type    = _running.Type    = (MidiEventType) (tmp & 240);
+                    _event.Channel = _running.Channel = (byte) (tmp & CHANNEL_MASK);
+                    _event.Type    = _running.Type    = (MidiEventType) (tmp & EVENTTYPE_MASK);
                     _event.Length  = _running.Length  = _running.Type switch
                     {
                         MidiEventType.Note_On => 2,
@@ -157,10 +160,13 @@ namespace YARG.Core.IO
             note.velocity = span[_trackPos + 1];
         }
 
-        private const uint EXTENDED_VLQ_FLAG = 128;
-        private const uint VLQ_MASK = 127;
+        private const uint EXTENDED_VLQ_FLAG = 0x80;
+        private const uint VLQ_MASK = 0x7F;
         private const int  VLQ_SHIFT = 7;
         private const int  MAX_SHIFTCOUNT = 3;
+        /// <summary>
+        /// Represents the minimum value where a VLQ shift would be illegal
+        /// </summary>
         private const uint VLQ_SHIFTLIMIT = 1 << (VLQ_SHIFT * MAX_SHIFTCOUNT);
         private uint ReadVLQ()
         {
--- a/YARG.Core/IO/Midi/YARGMidiTrack.cs
+++ b/YARG.Core/IO/Midi/YARGMidiTrack.cs
@@ -51,6 +51,7 @@ namespace YARG.Core.IO
         private MidiEvent _running;
 
         private readonly byte[] _data;
+        private readonly ReadOnlyMemory<byte> memory;
         private int _trackPos;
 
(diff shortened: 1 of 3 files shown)


YARG.Core/Song/Preparsers/Midi/MidiPreparser.cs
```diff
@@ -13,26 +13,28 @@ public abstract class Midi_Preparser
...
+                        if (ParseNote_Off(track))
+                            return true;
+                        break;
+                    case MidiEventType.SysEx:
+                    case MidiEventType.SysEx_End:
+                        ParseSysEx(track.ExtractTextOrSysEx());
+                        break;
+                    case <= MidiEventType.Text_EnumLimit:
+                        ParseText(track.ExtractTextOrSysEx());
+                        break;
```
TheNathannator, line 36: Should probably make sure this won't attempt to read sequence number events as text events, as that's meta type 0. Niche edge-case, but better safe than sorry lol

YARG.Core/IO/Midi/YARGMidiTrack.cs
```diff
@@ -46,32 +46,52 @@ private struct MidiEvent
...
         }
 
-        private long _position;
+        private long _tickPosition;
         private MidiEvent _event;
         private MidiEvent _running;
 
-        private YARGBinaryReader _reader;
+        private readonly byte[] _data;
+        private readonly ReadOnlyMemory<byte> _memory;
```
TheNathannator, line 54: Would it be possible to avoid the `ReadOnlyMemory` and only use the `byte[]`, given the performance difference between the two?

Currently the only function of the `_data` field is to store the data returned by `stream.ReadBytes(count)`, but that isn't necessary since `_memory` will already store a reference to the buffer (as an opaque `object`) and keep it from being garbage collected. Either `_data` should be removed, or `_memory` should be replaced with position/length values which are used to construct a `Span` manually.

YARG.Core/IO/Midi/YARGMidiTrack.cs
```diff
@@ -81,12 +101,12 @@ public bool ParseEvent()
...
             {
-                _reader.Move_Unsafe(1);
+                _trackPos++;
                 if (type < MidiEventType.SysEx)
                 {
-                    _running.Channel = (byte) (tmp & 15);
-                    _running.Type = (MidiEventType) (tmp & 240);
-                    _running.Length = _running.Type switch
+                    _event.Channel = _running.Channel = (byte) (tmp & 15);
+                    _event.Type    = _running.Type    = (MidiEventType) (tmp & 240);
```
TheNathannator, line 108: Would be preferable to replace the decimal values here with hexadecimal ones (`15` -> `0x0F`, `240` -> `0xF0`), to make it clearer how the values are being masked.

YARG.Core/IO/Midi/YARGMidiTrack.cs
```diff
@@ -123,23 +142,64 @@ public bool ParseEvent()
...
-            note.value = _reader.ReadByte();
-            note.velocity = _reader.ReadByte();
+            var span = _memory.Span;
+            note.value = span[_trackPos];
+            note.velocity = span[_trackPos + 1];
         }
 
-        public void SkipEvent()
+        private const uint EXTENDED_VLQ_FLAG = 128;
+        private const uint VLQ_MASK = 127;
```
TheNathannator, line 164: Same here as with the MIDI event values, use hexadecimal to make it clearer which bits are being used.

YARG.Core/IO/Midi/YARGMidiTrack.cs
```diff
@@ -123,23 +142,64 @@ public bool ParseEvent()
...
+        private const int  VLQ_SHIFT = 7;
+        private const int  MAX_SHIFTCOUNT = 3;
+        private const uint VLQ_SHIFTLIMIT = 1 << (VLQ_SHIFT * MAX_SHIFTCOUNT);
```
TheNathannator, line 167: This constant is a little confusing, since VLQ values can store up to 28 bits. I almost thought it was incorrect lol

TheNathannator (approved):
These comments don't really fit the scope of this PR, just adding them here so I don't forget to mention them lol

YARG.Core/Song/Preparsers/Midi/MidiPreparser.cs
sonicfind: Good point

YARG.Core/IO/Midi/YARGMidiTrack.cs
sonicfind: "_memory will already store a reference to the buffer"

Word?

YARG.Core/IO/Midi/YARGMidiTrack.cs
sonicfind: Unfortunately, there's not much you can do with the naming of that variable in particular. I just added a summary comment to make the function for it clearer.

TheNathannator: Yeah, that should be good.

PR #116, opened 2023-11-27
Allow processing inputs and updating to the current time in a single engine update

TheNathannator:
PR'ing to allow discussion and deliberation before merging. Feeling unsure about this one for some reason lol

This change removes the need to check if inputs are queued before updating the engine, now there's a method which handles both inputs and updating to the given time. The rationale is to ensure there are no discrepancies between where the engine updates to and the actual input update time.

Since the API's changed, this will require a couple changes in the main YARG project.

### DIFF
--- a/ReplayAnalyzer/Analyzer.cs
+++ b/ReplayAnalyzer/Analyzer.cs
@@ -121,11 +121,11 @@ public class Analyzer
                 // Run!
                 if (engine.IsInputQueued)
                 {
-                    engine.UpdateEngine();
+                    engine.UpdateEngineInputs();
                 }
                 else
                 {
-                    engine.UpdateEngine(frameTime);
+                    engine.UpdateEngineToTime(frameTime);
                 }
             }
         }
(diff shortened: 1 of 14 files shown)


RileyTheFox:
I was working on refactoring the update methods and using a new approach to resolve issues I realised could happen when updating the engine to a time.

image

I never pushed these changes from my local branch though so its my fault nobody knew what I'd done lol. I'm going to push it now and then you can take a look.

TheNathannator:
As I mentioned in the Discord:

The changes in the new branch don't really address the problem this PR is targeting, that being that there's unnecessary complexity when going to update the engine, and the current handling of that complexity causes a time disparity when inputs are queued (however minor it may be).

Ideally, it would just be a single update method which takes in the current time, and that method would handle both the queued inputs and the time update. Having separate methods for input-only and time-only updates is good, but those should only have to be used if they're necessary. Otherwise, the engine handles it on its own as an implementation detail.

I can understand doing it the current way to make it clearer as to what's going on (as you mentioned in the Discord as well), but IMO that's a job for the documentation to point out. Making the API more complex just to make the inner workings a little easier to see is not a great practice lol

RileyTheFox:
Could you possibly retarget this PR to the `engine-remake` branch as there's additional changes over there and it will be merged to master once it's finished

TheNathannator:
Done

TheNathannator:
Gonna close since this PR is effectively obsolete by this point lol

PR #117, opened 2023-11-27
Bass multiplier

grishhung:
Super quick and dirty reimplementation of the bass multiplier. Needed to be reimplemented on the engine side before I could do other changes related to text notifications and bass groove.

### DIFF
--- a/YARG.Core/Chart/Tracks/InstrumentDifficulty.cs
+++ b/YARG.Core/Chart/Tracks/InstrumentDifficulty.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using YARG.Core.Extensions;
 
--- a/YARG.Core/Engine/Guitar/GuitarEngine.cs
+++ b/YARG.Core/Engine/Guitar/GuitarEngine.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using YARG.Core.Chart;
 using YARG.Core.Engine.Logging;


YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -23,12 +23,16 @@ public abstract class GuitarEngine : BaseEngine<GuitarNote, GuitarAction, Guitar
 
         public override bool TreatChordAsSeparate => false;
 
+        private bool _isBass;
```
RileyTheFox, line 26: This should be a guitar engine parameter passed in from YARG

YARG.Core/Chart/Tracks/InstrumentDifficulty.cs
```diff
@@ -87,6 +87,14 @@ public bool IsOccupied()
             return Notes.Count > 0 || Phrases.Count > 0 || TextEvents.Count > 0;
         }
 
+        public bool IsBass()
```
RileyTheFox, line 90: Should be handled on the YARG side using the player's profile instrument. Having a function in here that is somewhat instrument specific violates its generic-ness.

TheNathannator (changes requested):
Try to avoid "quick and dirty" if you can help it, that's kinda why we rewrote everything in the first place lol

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -228,13 +232,28 @@ protected override void AddScore(GuitarNote note)
...
         {
-            EngineStats.ScoreMultiplier = EngineStats.Combo switch
+            if (_isBass)
             {
-                >= 30 => 4,
-                >= 20 => 3,
-                >= 10 => 2,
-                _     => 1
-            };
+                EngineStats.ScoreMultiplier = EngineStats.Combo switch
```
RileyTheFox, line 237: The _isBass class field should be removed and an engine parameter be opted for instead.

This if/else switch expression can be de-duplicated by doing something like the following:

```
var isBass = EngineParameters.IsBass;

EngineStats.ScoreMultiplier = EngineStats.Combo switch
{
    >= 50 => isBass ? 6 : 4,
    >= 40 => isBass ? 5 : 4,
    >= 30 => 4,
    >= 20 => 3,
    >= 10 => 2,
    _     => 1
};
```

RileyTheFox (changes requested):

TheNathannator:
lol, we're both on top of this right now

EliteAsian123:
Closing as this was added in https://github.com/YARC-Official/YARG.Core/commit/8009190f38772a93c0d3dcd8ca31072847e76567

PR #120, opened 2023-12-04
Disallow .mid files with multiple track name events

TheNathannator:
Because these files just needed *yet another* edge-case for us to handle...

### DIFF
--- a/YARG.Core/Extensions/MidiExtensions.cs
+++ b/YARG.Core/Extensions/MidiExtensions.cs
@@ -32,12 +32,17 @@ namespace YARG.Core.Extensions
 
         public static string GetTrackName(this TrackChunk track)
         {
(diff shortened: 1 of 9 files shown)


sonicfind:
Gonna add some changes to the YARGMidiTrack function to have it return the trackname string instead of the trackType.

sonicfind:
Because we can do that, lol

TheNathannator:
Your changes to `FindTrackName` defeat the entire purpose of the PR lol, now it once again only takes the first track name in the file.

TheNathannator:
Not too enthused about this PR turning into the opposite of what it was originally supposed to be, but since everyone else seems to be against allowing multiple track names, I guess there's not much I can argue against.

PR #121, opened 2023-12-10
Move over audio interfaces and pitch detection code, along with reworks and fixes

TheNathannator:
Draft while I evaluate these changes throughout implementing them in main YARG.

The audio interfaces have been moved over, which allows more code to be brought over to Core, such as the audio loading extensions (moved in this PR) and the song runner (to be moved in the future). In addition, the pitch detection code has been moved over, and its original license/readme have been tracked down and added to its directory.

A number of reworks/cleanups have happened as well, which will require changes in main YARG. This includes:

- Audio loading is now inherently asynchronous, rather than having to manually wrap it in `Task.Run` or `UniTask.RunOnThreadPool`.
- Things related to song playback have been moved to a new `ISongContext` interface: loading a song will return an instance of this interface, and all stats/control of the song playback is done through that instead of through the audio manager.
  - This should help prevent odd concurrency issues that can cause the audio backend to explode, primarily by making responsibility a little more clear, and containing the mixer/channels for a song to a dedicated object rather than using only a single effectively-global mixer.
- Various other changes for things that I didn't think made a whole lot of sense, such as methods being used in place of property getters/setters, and other design choices I thought could be simplified/reworked.
- SFX stuff has been split out into its own interface and made generic, so that Core doesn't have to carry project-specific SFX definitions.
  - This interface is also not tied to `IAudioManager`, to allow a bit more flexibility in how it is used.


TheNathannator:
Closing since this has been superseded by sonicfind's audio rework.

PR #123, opened 2023-12-19
Yet Another Moonscraper Refactor (YAMR), plus parsing unit test fixes

TheNathannator:
Lots of cleanup for the `SongObject` family: redundant types have been removed/merged, equality/comparison has been reworked into separate value-equality and insertion-comparison semantics, and other things that bugged me have been adjusted. The event types have also been given a `Moon` prefix to avoid further confusion between Moonscraper's events and `YARG.Core.Chart`'s.

Many things with the parsing behavior unit tests broke in the process of both this and the direct insertion PR, and have now been fixed.


YARG.Core/MoonscraperChartParser/Events/MoonNote.cs
```diff
@@ -303,33 +290,25 @@ public int GetMaskWithRequiredFlags(Flags flags)
...
+                        ? MoonNoteType.Tap
+                        : IsHopo(hopoThreshold) ? MoonNoteType.Hopo : MoonNoteType.Strum,
+
+                MoonChart.GameMode.Drums =>
+                    drumPad is DrumPad.Yellow or DrumPad.Blue or DrumPad.Orange && (flags & Flags.ProDrums_Cymbal) != 0
+                        ? MoonNoteType.Cymbal
+                        : MoonNoteType.Strum,
+                _ => MoonNoteType.Natural
+            };
         }
```
sonicfind, line 310: Not entirely sure why this specific change was necessary. The contexts that called either of these already very explicitly answer the question of "Is this drums or guitar?". Combining them into one requires 1. an inherent loss in the communication of some logical certainties, & 2. requires sending additional information that will go unused depending on which you're calling for.

Drums don't require guitar-related information for its query (nor the GameMode), but the parameters now force callers to pass those in for some reason.

**Side note: I'm not a fan of the nested ternary style. It just ends up being harder to read.**

TheNathannator: This was a compromise in order to ensure the assert in MidReader's NoteOnAsForcedType behaves correctly on all game modes without having to manually check the game mode first. I know it's not used on drums, but it was the easiest way of making it work correctly without making that assumption.

YARG.Core/MoonscraperChartParser/IO/Midi/MidReader.cs
```diff
@@ -667,10 +667,11 @@ private static void ProcessEventAsForcedTypePostDelay(in EventProcessParams even
         {
             var song = eventProcessParams.song;
             var instrument = eventProcessParams.instrument;
+            var gameMode = MoonSong.InstumentToChartGameMode(instrument);
```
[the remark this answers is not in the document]
TheNathannator: Wasn't aware that was a thing for some reason lol

YARG.Core/MoonscraperChartParser/Events/MoonNote.cs
sonicfind: > This was a compromise in order to ensure the assert in MidReader's NoteOnAsForcedType behaves correctly on all game modes

The problem here is that `NoteOnAsForcedType` is only ever used in guitar-related contexts, If you step up to all the functions that'll call it (and to the functions that call those), they always lead back to the guitar-based processList collections. Therefore, "all the game modes" in this situation is actually "all [guitar] game modes", which was already handled.

It's not that it's an *assumption* of it being a guitar mode, it's that the logic *guaranteed* that it is a guitar mode.

TheNathannator: Yes, that is why I mention this:

> we probably should make that check up-front instead of going through the marking anyways on drums

Otherwise there's nothing statically guaranteeing that we will be on guitar, and it's just an assumption we're making about how the function is used.

That first paragraph is the thought process I had at the time when making the change, mainly just explaining myself lol

PR #128, opened 2023-12-31
Auto-Generate Sections When Songs Don't Have Any

EliteAsian123:
When a song is loaded without any sections, 10 equal sections are created for the song automatically. I wasn't too sure exactly where to put this, since we need the last tick of the chart, so I just made a  `PostProcessSections` method in `SongChart`, and called it at the end of the constructor where everything is loaded. Not sure if that's the best place, let me know if there's a better one.

Also moved the code ensuring that the last section extends to the end of the chart from YARG to here. Small YARG change required when this is merged because of it.

### DIFF
--- a/YARG.Core/Chart/SongChart.cs
+++ b/YARG.Core/Chart/SongChart.cs
@@ -163,7 +163,10 @@ namespace YARG.Core.Chart
                     double startTime = SyncTrack.TickToTime(startTick);
                     double endTime   = SyncTrack.TickToTime(endTick);
 
-                    var section = new Section($"Section {i + 1}", startTime, startTick)
+                    // 10%, 20%, 30%, etc.
+                    var sectionName = $"{100f / AUTO_GEN_SECTION_COUNT * (i + 1):N0}%";
+
+                    var section = new Section(sectionName, startTime, startTick)
                     {
                         TickLength = endTick - startTick,
                         TimeLength = endTime - startTime,
--- a/YARG.Core/Chart/SongChart.cs
+++ b/YARG.Core/Chart/SongChart.cs
@@ -155,16 +155,18 @@ namespace YARG.Core.Chart
             // This prevents issues with songs with no sections, such as in practice mode.
             if (Sections.Count <= 0)
             {
-                for (uint i = 0; i < AUTO_GEN_SECTION_COUNT; i++)
+                uint startTick = 0;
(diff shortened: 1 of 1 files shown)


YARG.Core/Chart/SongChart.cs
```diff
@@ -141,6 +141,47 @@ internal SongChart(ISongLoader loader)
...
+                    var sectionName = $"{100f / AUTO_GEN_SECTION_COUNT * (i + 1):N0}%";
+
+                    var section = new Section(sectionName, startTime, startTick)
+                    {
+                        TickLength = endTick - startTick,
+                        TimeLength = endTime - startTime,
+                    };
+
+                    Sections.Add(section);
+                }
```
sonicfind, line 176: Use an array of 11 uints. Fill the array with the TickToTime results so that you don't have to recalculate them in each iteration, with the last one being lastTick.

EliteAsian123: I'll do you one better: no array needed!

YARG.Core/Chart/SongChart.cs
```diff
@@ -155,16 +155,18 @@ private void PostProcessSections()
...
+                // always one index behind.
+                for (uint i = 1; i <= AUTO_GEN_SECTION_COUNT; i++)
+                {
+                    uint endTick = lastTick / AUTO_GEN_SECTION_COUNT * i;
+                    double endTime = SyncTrack.TickToTime(endTick);
 
-                    // 10%, 20%, 30%, etc.
-                    var sectionName = $"{100f / AUTO_GEN_SECTION_COUNT * (i + 1):N0}%";
+                    // 10%, 20%, 30%, etc. (the ending percent)
+                    var sectionName = $"{100f / AUTO_GEN_SECTION_COUNT * i:N0}%";
```
sonicfind, line 169: The percentage should be where the section *starts*. Doesn't make sense to have the first section be "10%" when you actually start at "0%".

EliteAsian123: I agree, that's just what the folks on Discord told me. I'm honestly just gonna change it to “10% to 20%”, screw what CH does lol

sonicfind: .... which is that

EliteAsian123: I WAS FOOLED

PR #136, opened 2024-01-14
Implement guitar sustain scoring

TheNathannator:
The long-awaited return, now (hopefully) 100% consistent!

This uses very similar principles to the Star Power gain/drain rework, and shares the same infrastructure. I've done testing to make sure all the edge cases I can think of work correctly, but some more stress testing should be done to make sure nothing was missed. I'll be doing some bot runs between YARG and CH to compare scores.

### DIFF
--- a/YARG.Core/Game/Presets/EnginePreset.Instruments.cs
+++ b/YARG.Core/Game/Presets/EnginePreset.Instruments.cs
@@ -7,7 +7,6 @@ namespace YARG.Core.Game
     public partial class EnginePreset
     {
         public const double DEFAULT_WHAMMY_BUFFER = 0.25;
-        public const double DEFAULT_SUSTAIN_BURST = 0.075;
 
         public const int DEFAULT_MAX_MULTIPLIER = 4;
         public const int BASS_MAX_MULTIPLIER    = 6;
@@ -76,7 +75,6 @@ namespace YARG.Core.Game
                     StrumLeniency,
                     StrumLeniencySmall,
                     DEFAULT_WHAMMY_BUFFER,
-                    DEFAULT_SUSTAIN_BURST,
                     InfiniteFrontEnd,
                     AntiGhosting);
             }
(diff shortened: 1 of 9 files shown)


YARG.Core/Game/Presets/EnginePreset.Instruments.cs
```diff
@@ -7,6 +7,7 @@ namespace YARG.Core.Game
     public partial class EnginePreset
     {
         public const double DEFAULT_WHAMMY_BUFFER = 0.25;
+        public const double DEFAULT_SUSTAIN_BURST = 0.075;
```
RileyTheFox, line 10: Sustain burst window isn't a pre-determined time. It tends to be done based on current beat times, usually 1/4 of a beat.

I think a variable time is better compared to a static one

TheNathannator: I personally think it makes more sense to do it based on time, that way the leniency is consistent and explicitly *doesn't* depend on tempo. It would also make the option easier to comprehend were it to become configurable in the future.

RileyTheFox: That's a good point but the reason for it being dynamic is that it works for all tempos. At really high tempo, you'd have to hold it closer to the complete end which makes it a bit harder to transition from one sustain to the next (and at high tempo, you want that because faster should be somewhat harder)

And at low tempo, the song is typically more relaxed, so the leniency is higher

TheNathannator:
Drafting for now as I notice more problems that need to be worked out. Scores aren't 100% consistent for some reason, and base score calculation doesn't match how the actual points are calculated during gameplay.

TheNathannator:
Scoring inconsistencies fixed, and things should match the base score now.

YARG.Core/Game/Presets/EnginePreset.Instruments.cs
TheNathannator: Changed to ticks after SP pathing/squeezing was brought up while discussing in the Discord server.

EliteAsian123 (approved):

RileyTheFox (approved):

PR #138, opened 2024-01-23
Fix some bugs and edgecases in NoKick mode

jgoyvaerts:


EliteAsian123:
Getting a ton of `Engine was forced to move an input time!` on vocals

PR #145, opened 2024-02-09
Place unterminated events at the end of the song, fixing song length

jgoyvaerts:


YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Lyrics.cs
```diff
@@ -68,7 +69,8 @@ public void AddPhraseEvent(string text, uint tick)
         public LyricsTrack LoadLyrics()
         {
             var converter = new LyricConverter(_moonSong);
-            TextEvents.ConvertToPhrases(_moonSong.events, converter);
+            var maxTick = _moonSong.Charts.Max(x => x.notes.LastOrDefault()?.tick ?? 0);
```
TheNathannator, line 72: This needs to account for the lyric events as well, not just the notes. Probably better to use `events` here rather than `Charts`/`notes`.

Also would prefer explicitly accessing the list rather than using Linq, as Linq operates exclusively with `IEnumerable<T>` and will iterate over the entire list here to find the last item (assuming there isn't a specialization fast path for `List<T>` in the implementation).

```diff
- var maxTick = _moonSong.Charts.Max(x => x.notes.LastOrDefault()?.tick ?? 0);
+ int maxTick = _moonSong.events.Count > 0 ? _moonSong.events[^1] : 0;
```

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.ProcessLists.cs
```diff
@@ -229,7 +229,8 @@ public void AddPhrase(uint startTick, uint endTick)
         private static void ConvertSoloEvents(in NoteProcessParams noteProcessParams)
         {
             var chart = noteProcessParams.chart;
-            TextEvents.ConvertToPhrases(chart.events, new SoloPhraseConverter(chart));
+            var maxTick = chart.notes.LastOrDefault()?.tick ?? 0;
```
TheNathannator, line 232: Same preference here:

```diff
- var maxTick = chart.notes.LastOrDefault()?.tick ?? 0;
+ int maxTick = chart.notes.Count > 0 ? chart.notes[^1] : 0;
```

TheNathannator (changes requested):

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Lyrics.cs
jgoyvaerts: Added the resolution and changed to events as requested.

No need to switch from Linq to direct: https://github.com/mono/corefx/blob/c4eeab9fc2faa0195a812e552cd73ee298d39386/src/System.Linq/src/System/Linq/Last.cs#L51

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.ProcessLists.cs
jgoyvaerts: No need to switch from Linq to direct: https://github.com/mono/corefx/blob/c4eeab9fc2faa0195a812e552cd73ee298d39386/src/System.Linq/src/System/Linq/Last.cs#L51

TheNathannator: We determined there is an optimization path for lists, so this is not much of a concern anymore:
https://github.com/mono/corefx/blob/5b4edb9b0a721aed60c9c8ddeed441bdd7451e42/src/System.Linq/src/System/Linq/Last.cs#L51

TheNathannator (approved):

PR #149, opened 2024-02-19
Replace SongMetadata With An Inheritance Tree Of SongEntry Types

sonicfind:
Provides a more streamlined interface for accessing and loading information/data from song entries without needing to know the inner workings of the *actual type* of the entry.

There's the base abstract type that first splits between `Ini` and `RBCON` - both of which take a final split into `Packed` and `Unpacked`. Should be pretty straight forward.

Additionally, we consolidate the metadata held by entries within separate composed structures: one for shared base information and another for extended RBCON information. This opens the doors for tools that allow easily editing those values in a separate context.

This refactor makes it easier to support a bunch of different features we didn't have before. For example, we can directly utilize RBCON names within an entry to load custom backgrounds and such.

+ Heavily requires an accompanying main repo update to work properly


EliteAsian123 (approved):
Did a quick skim, I don't see anything glaringly wrong, and I think we should do this anyway as it's a good idea. Feel free to merge yourself so the YARG side PR can be merged at the same time.

PR #150, opened 2024-02-20
Convert ParseSettings To A Struct

sonicfind:
Requires usage of `ref` during song loading to keep modifications valid.


TheNathannator (approved):
On one hand, needing `ref` everywhere is precisely the reason I made it a class to begin with, it felt rather error-prone just to allow modifications or avoid redundant copies.

On the other hand however, taking notice of commit [`1614949`](https://github.com/YARC-Official/YARG.Core/commit/1615949efedcf08c55479a374dacca6a89e63dd9), the possibility of different song entries unintentionally using the same settings instance is a problem of its own caliber which I'm happy to take into account here.

PR #155, opened 2024-03-17
Fully consistent guitar engine (not counting sustains and starpower)

EliteAsian123:
Lots of testing later, it's finally here (yippe!)

**Firstly, caveats and things I'd like to improve:**
* This reverts to the old timer system, as the replacement of the `IsActive` method with the `IsEnabled` property does not provide consistent results, as the current time is not taken into consideration, and is therefore not properly disabled on time. I may be misunderstanding something with it; I didn't spend too long trying to fix it.
* `InfiniteFrontEndHitTime` is not a timer, and probably should be.
* The current consistency anchor system is slow, as there are a lot of updates queued that are actually not needed most of the time.
* `RunHitLogic` has been reverted to a loop. It should theoretically be possible to keep it as one update call, but I'm yet to figure out why removing the loop causes inconsistency problems.
* A recursive call is used for note skipping. I'm not sure if this is the best way of doing it, but it's certainly the easiest.

All note hitting logic along with all of its leniencies are fully consistent as far as I can tell. The only consistency issues seem to come from sustain points, and star power ending slightly earlier or later than it should.

I've done extensive testing with comparing this version of the engine with the master branch one, and compared inconsistencies between them. All inconsistencies between the engines could be justified properly and ruled out (i.e., fast strumming sections dropping random notes when the strum was clearly present: hit on this engine, not on the old one). Big disclaimer though: I am really bad, so there could be some minor vicissitudes that I have not noticed. Let me know, and these can easily be fixed.

This is a complete engine reimplementation that uses a different ideology (if you will) which is a lot easier to follow through (in my opinion at least). It's a lot easier to visualize the flow structure of everything, which I have discovered is very important for consistency related problems, as it becomes *very* clear what the potential problems are.

P.S. I've also just realized that HOPO leniency should *probably* be reset after a note gets hit. I'll add this on later when I get the chance, if it is actually the case.

### DIFF
--- a/YARG.Core/Engine/BaseEngine.Generic.cs
+++ b/YARG.Core/Engine/BaseEngine.Generic.cs
@@ -18,16 +18,16 @@ namespace YARG.Core.Engine
 
         // Max number of measures that SP will last when draining
         // SP draining is done based on measures
-        protected const int STAR_POWER_MAX_MEASURES = 8;
+        protected const int    STAR_POWER_MAX_MEASURES   = 8;
         protected const double STAR_POWER_MEASURE_AMOUNT = 1.0 / STAR_POWER_MAX_MEASURES;
 
         // Max number of beats that it takes to fill SP when gaining
         // SP gain from whammying is done based on beats
-        protected const int STAR_POWER_MAX_BEATS = (STAR_POWER_MAX_MEASURES * 4) - 2; // - 2 for leniency
+        protected const int    STAR_POWER_MAX_BEATS   = (STAR_POWER_MAX_MEASURES * 4) - 2; // - 2 for leniency
         protected const double STAR_POWER_BEAT_AMOUNT = 1.0 / STAR_POWER_MAX_BEATS;
 
         // Number of measures that SP phrases will grant when hit
-        protected const int STAR_POWER_PHRASE_MEASURE_COUNT = 2;
+        protected const int    STAR_POWER_PHRASE_MEASURE_COUNT = 2;
         protected const double STAR_POWER_PHRASE_AMOUNT = STAR_POWER_PHRASE_MEASURE_COUNT * STAR_POWER_MEASURE_AMOUNT;
 
         // Beat fraction to use for the sustain burst threshold
@@ -57,9 +57,9 @@ namespace YARG.Core.Engine
         public SoloStartEvent? OnSoloStart;
         public SoloEndEvent?   OnSoloEnd;
 
-        protected int[] StarScoreThresholds { get; }
+        protected          int[]  StarScoreThresholds { get; }
         protected readonly double TicksPerSustainPoint;
-        protected readonly uint SustainBurstThreshold;
+        protected readonly uint   SustainBurstThreshold;
 
         public readonly TEngineStats EngineStats;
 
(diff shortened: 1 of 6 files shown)


RileyTheFox (changes requested):
In my minimal testing the engine is producing consistent results, however I have a number of problems with this approach:

- The hit logic is written around a "consistency" approach instead of just actual hit logic which makes it hard to change logic or add new things in the future as you have to consider if its a "consistent" method.
- `UpdateEngineLogic` is almost twice the length as the previous version when its not actually doing a whole lot differently, really hard to follow
- The way infinite front end is checked is confusing
- `ProcessNote` is called in UpdateEngineLogic in about 4 different places, and then is also recursively called within itself. Its extremely hard for me to figure out what conditions are causing what from a glance

Infinite front end is also not functioning correctly after strumming:

https://github.com/YARC-Official/YARG.Core/assets/29326764/e9b8fefd-5b30-4ff8-9519-59f251258f17

I think the "consistency" part of the engine should be placed on the caller of the hit logic (so the functions in BaseEngine) and not the callee (UpdateEngineLogic). I think it would be much simpler if the caller in BaseEngine is responsible for generating update times in between 2 frames/times and only the hit logic is concerned about in each engine's code.

EliteAsian123:
> The hit logic is written around a "consistency" approach instead of just actual hit logic which makes it hard to change logic or add new things in the future as you have to consider if its a "consistent" method.

This is not a problem with this implementation and is just a problem with the frame independent approach in general. For example, in another attempted rewrite of the engine, everything was working fine until note skipping was added, and this was using something similar to the previous implementation. 

> UpdateEngineLogic is almost twice the length as the previous version when its not actually doing a whole lot differently, really hard to follow

This is because the stuff from `UpdateTimers`, `HitNote`, and `MissNote` have all been moved into the `UpdateEngineLogic` method instead of separating it. It's actually around the same length, or shorter if you count everything else in the file (however, I don't think that's a fair comparison).

I'd argue that because of this, it's quite a bit easier to follow, especially for the `HitNote` and `MissNote` methods.

> The way infinite front end is checked is confusing

I believe (I think?) that we discussed this before, but it is a little more confusing since you need to calculate when it will hit. That's the sanest way to do it consistently. 

> ProcessNote is called in UpdateEngineLogic in about 4 different places, and then is also recursively called within itself. Its extremely hard for me to figure out what conditions are causing what from a glance

`ProcessNote` is basically the new `CheckForNoteHit`, with arguments added on. In my opinion, it is actually way easier to understand than before because *multiple* different types of inputs and leniencies cause note hits, not just one (in fact, it's 4: infinite front end, strumming, tapping, strum leniency). 

> Infinite front end is also not functioning correctly after strumming

It looks like the second call to `CheckInfiniteFrontEndAndGhost` was removed at one point. Not sure how that happened. I'll fix it after I comment this. Sorry, good catch!

> I think the "consistency" part of the engine should be placed on the caller of the hit logic (so the functions in BaseEngine) and not the callee (UpdateEngineLogic). I think it would be much simpler if the caller in BaseEngine is responsible for generating update times in between 2 frames/times and only the hit logic is concerned about in each engine's code.

If I'm understanding what you're saying, this is unrealistic. Because of how engine consistency works, *every little detail* must be accounted for, and that is one of the things I've learned through creating this version of the engine. An engine is a complex system. You cannot simply implement something is the “most intuitive” or “best for gameplay.” Because of how fragile consistency is, you must consider that before anything, as gameplay isn't as fragile. Apologies if I misunderstood what you meant.

EliteAsian123:
Fixed. I didn't do much testing at all, I can do thorough testing tomorrow night.

EliteAsian123:
Closing as a different approach is now being used in `the-final-rewrite` branch

PR #159, opened 2024-04-07
Added None as a LightingType

TheFatBastid:
There is quite a few checks against CurrentLightingCue being null, this will allow it to always be set to something, removing costly null checks.


TheNathannator:
FWIW nullable value types are still value types and don't do any boxing to the heap, null checks on them are no more expensive than a single boolean check. Having this as a proper state is still desirable if it's something that should be generally allowed, however.

I'd like to see this change applied to places before this is merged, feels pointless to merge something if it doesn't do anything lol

TheFatBastid:
Obsoleted by upcoming external lighting program

PR #164, opened 2024-05-05
FILETABLEBLOCKCOUNT wrong endianness

luis-mejia1:
The `FILETABLEBLOCKCOUNT` of CON files is stored as a little-endian int16, but it is being read as a big-endian number.

This causes the following error when loading some multi-song CON files, like the part 1 of the `Guitar Hero Series Main` pack from RB3DX's spreadsheet.

> [Error] [2024-05-04 22:41:05 CONFile.cs:TryParseListings:87] Error while parsing *\yarg\songs\gh\GHRB3main01 - Filelisting blocks constructed out of spec

### DIFF
--- a/YARG.Core/IO/ConHandler/CONFile.cs
+++ b/YARG.Core/IO/ConHandler/CONFile.cs
@@ -61,7 +61,7 @@ namespace YARG.Core.IO
             if (stream.Read(int32Buffer[..BYTES_16BIT]) != BYTES_16BIT)
                 return null;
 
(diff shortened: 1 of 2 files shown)


sonicfind:
If you could also do the same for the numBlocks value in CONFileLisining, that'd be great

luis-mejia1:
> If you could also do the same for the numBlocks value in CONFileListing, that'd be great

Absolutely, I've made the same changes for the CONFileListing numBlocks.

PR #167, opened 2024-05-10
Cleanup TextContainer Usage

sonicfind:
Converts YARGChartFileReader & YARGTextReader to static classes

### DIFF
--- a/YARG.Core/IO/Ini/IniModifierCreator.cs
+++ b/YARG.Core/IO/Ini/IniModifierCreator.cs
@@ -38,9 +38,8 @@ namespace YARG.Core.IO.Ini
             this.type = type;
         }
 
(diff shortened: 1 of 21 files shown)


TheNathannator (commented):
I'm not sure if I'm a fan of everything being converted to a static class and a struct, what's the reasoning behind it?

Having to pass everything by `ref` everywhere seems clunky and error-prone if one happens to be missed in a parameter somewhere, as it'll result in invisible misbehavior due to the text container being copied rather than referenced. Using a class instance instead makes this issue impossible.

When ParseSettings was converted to a struct, that was fine as it's not used in many places and it doesn't have much functionality; it's mainly just a data structure. With this, you're slowly unravelling classes that *do* have significant behavior, and it's affecting many places in the code. Perhaps there's a better way to approach things if this is being done for performance reasons.

PR #169, opened 2024-05-14
Wait countdown core

sunset-cf:
Added hooks into YARG.Core to check for stretches with no notes in a chart


sunset-cf:
Sorry, a lot of auto-formatting changes went through. I will take those out and re-request

PR #170, opened 2024-05-14
YARG.Core changes for WaitCountdown support

Purplo-cf:
Additions to the Guitar, Drums, and Vocals engines to support countdowns during long stretches of no notes

### DIFF
--- a/YARG.Core/Chart/Events/WaitCountdown.cs
+++ b/YARG.Core/Chart/Events/WaitCountdown.cs
@@ -46,7 +46,7 @@ namespace YARG.Core.Chart
             {
                 newMeasuresLeft = 0;
             }
-            else if (currentTick == Tick)
+            else if (currentTick < Tick)
             {
                 newMeasuresLeft = TotalMeasures;
             }
--- a/YARG.Core/Chart/Events/WaitCountdown.cs
+++ b/YARG.Core/Chart/Events/WaitCountdown.cs
@@ -5,7 +5,7 @@ namespace YARG.Core.Chart
 {
     public class WaitCountdown : ChartEvent
     {
-        public const float MIN_SECONDS = 10;
+        public const float MIN_SECONDS = 9;
         public const uint MIN_MEASURES = 4;
         public const float MIN_UPDATE_SECONDS = 1;
         private const float MIN_GET_READY_SECONDS = 2;
--- a/YARG.Core/Chart/Events/WaitCountdown.cs
+++ b/YARG.Core/Chart/Events/WaitCountdown.cs
@@ -39,25 +39,33 @@ namespace YARG.Core.Chart
             _measuresLeft = TotalMeasures;
         }
 
-        public int GetRemainingMeasures(uint currentTick)
+        public int GetRemainingMeasures(uint currentTick, out bool needsUpdate)
         {
-            int measuresLeft;
+            int newMeasuresLeft;
             if (currentTick >= TickEnd)
             {
-                measuresLeft = 0;
+                newMeasuresLeft = 0;
             }
             else if (currentTick == Tick)
             {
-                measuresLeft = TotalMeasures;
+                newMeasuresLeft = TotalMeasures;
             }
             else
             {
-                measuresLeft = TotalMeasures - _measureBeatlines.GetIndexOfNext(currentTick);
+                newMeasuresLeft = TotalMeasures - _measureBeatlines.GetIndexOfNext(currentTick);
             }
 
-            _measuresLeft = measuresLeft;
+            if (newMeasuresLeft != _measuresLeft)
+            {
+                _measuresLeft = newMeasuresLeft;
+                needsUpdate = true;
+            }
+            else
+            {
+                needsUpdate = false;
+            }
 
-            return measuresLeft;
+            return newMeasuresLeft;
         }
 
         public double GetNextUpdateTime()
--- a/YARG.Core/Chart/Events/WaitCountdown.cs
+++ b/YARG.Core/Chart/Events/WaitCountdown.cs
@@ -1,0 +1,184 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.NetworkInformation;
+
+namespace YARG.Core.Chart
+{
+    public class WaitCountdown : ChartEvent
+    {
+        public const double MIN_SECONDS = 10;
+        public const uint MIN_MEASURES = 4;
+        public const uint GET_READY_MEASURE = 2;
+        public const uint END_COUNTDOWN_MEASURE = 1;
+        private const double MIN_GET_READY_SECONDS = 2;
+
+        public readonly uint TotalMeasures;
+
+        public readonly double GetReadyTime;
+
+        private List<CountdownTimeSig> _measuresByTimeSignature;
+
+        private uint _measuresLeft;
+
+        public static uint NormalizeMeasures(uint providedMeasures, TimeSignatureChange sig)
+        {
+            return (uint) (providedMeasures / GetMeasureNormalizeMultiplier(sig));
+        }
+
+        public static uint GetPerNormalizedMeasure(uint providedTicks, TimeSignatureChange sig)
+        {
+            return (uint) (providedTicks / GetMeasureNormalizeMultiplier(sig));
+        }
+
+        public static double GetPerNormalizedMeasure(double providedSeconds, TimeSignatureChange sig)
+        {
+            return providedSeconds / GetMeasureNormalizeMultiplier(sig);
+        }
+
+        private static double GetMeasureNormalizeMultiplier(TimeSignatureChange sig)
+        {
+            double multiplier = 1;
+
+            // e.g. Normalize 2/4 and 1/4 to 4/4
+            if (sig.Numerator <= sig.Denominator / 2)
+            {
+                multiplier = sig.Denominator / sig.Numerator;
+            }
+
+            return multiplier;
+        }
+
+        public WaitCountdown(List<CountdownTimeSig> measuresByTimeSignature, uint totalMeasures=0)
+        {
+            _measuresByTimeSignature = measuresByTimeSignature;
+
+            var lastTimeSignature = measuresByTimeSignature.Last();
+
+            Time = measuresByTimeSignature[0].Time;
+            // Length properties should reflect the time the countdown spends onscreen, not the total time spent waiting
+            double timeEnd = lastTimeSignature.Time + lastTimeSignature.SecondsPerMeasure * (lastTimeSignature.TotalMeasures - END_COUNTDOWN_MEASURE);
+            TimeLength = timeEnd - Time;
+
+            Tick = measuresByTimeSignature[0].Tick;
+            uint tickEnd = lastTimeSignature.Tick + lastTimeSignature.TicksPerMeasure * (lastTimeSignature.TotalMeasures - END_COUNTDOWN_MEASURE);
+            TickLength = tickEnd - Tick;
+
+            if (totalMeasures == 0)
+            {
+                TotalMeasures = (uint) _measuresByTimeSignature.Sum(a => a.TotalMeasures);
+            }
+            else
+            {
+                TotalMeasures = totalMeasures;
+            }
+
+            _measuresLeft = TotalMeasures;
+
+            double getReadyTotalSeconds = 0;
+            uint measuresRemoved = 0;
+            int i = _measuresByTimeSignature.Count - 1;
+            while (measuresRemoved < GET_READY_MEASURE)
+            {
+                var currentStoredSig = _measuresByTimeSignature[i];
+                
+                uint maxMeasuresToRemove = GET_READY_MEASURE - measuresRemoved;
+                uint currentMeasuresToRemove;
+                if (currentStoredSig.TotalMeasures < maxMeasuresToRemove)
+                {
+                    currentMeasuresToRemove = maxMeasuresToRemove - currentStoredSig.TotalMeasures;
+                }
+                else
+                {
+                    currentMeasuresToRemove = maxMeasuresToRemove;
+                }
+
+                getReadyTotalSeconds += currentMeasuresToRemove * currentStoredSig.SecondsPerMeasure;
+
+                measuresRemoved += currentMeasuresToRemove;
+                i--;
+            }
+
+            GetReadyTime = TimeEnd - Math.Max(getReadyTotalSeconds, MIN_GET_READY_SECONDS);
+        }
+
+        public uint GetRemainingMeasures(uint currentTick)
+        {
+            if (currentTick >= TickEnd)
+            {
+                return 0;
+            }
+            
+            uint measuresLeft = TotalMeasures;
+            int totalTimeSignatures = _measuresByTimeSignature.Count;
+
+            for (int i = 0; i < totalTimeSignatures; i++)
+            {
+                var currentTimeSigReference = _measuresByTimeSignature[i];
+
+                if (currentTick < currentTimeSigReference.Tick)
+                {
+                    break;
+                }
+                else if (i < totalTimeSignatures-1 && currentTick > _measuresByTimeSignature[i+1].Tick)
+                {
+                    measuresLeft -= currentTimeSigReference.TotalMeasures;
+                    continue;
+                }
+
+                uint measuresElapsedThisSig = (currentTick - currentTimeSigReference.Tick) / currentTimeSigReference.TicksPerMeasure;
+                measuresLeft -= measuresElapsedThisSig;
+            }
+
+            _measuresLeft = measuresLeft;
+
+            return measuresLeft;
+        }
+
+        public double GetNextUpdateTime()
+        {
+            uint nextMeasureIndex = TotalMeasures - _measuresLeft + 1;
+            double nextUpdateTime = TimeEnd;
+            
+            // Countdown will be static during the "Get Ready!" window
+            if (nextMeasureIndex <= TotalMeasures - GET_READY_MEASURE)
+            {
+                uint runningMeasureTotal = 0;
+
+                for (int i = 0; i < _measuresByTimeSignature.Count; i++)
+                {
+                    var currentTimeSigReference = _measuresByTimeSignature[i];
+                    if (nextMeasureIndex > currentTimeSigReference.TotalMeasures + runningMeasureTotal)
+                    {
+                        runningMeasureTotal += currentTimeSigReference.TotalMeasures;
+                    }
+                    else
+                    {
+                        nextUpdateTime = currentTimeSigReference.Time + (currentTimeSigReference.SecondsPerMeasure * (nextMeasureIndex - runningMeasureTotal));
+                        break;
+                    }
+                }
+            }
+
+            return nextUpdateTime;
+        }
+    }
+
+    public struct CountdownTimeSig
+    {
+        public readonly uint Tick;
+        public readonly double Time;
+        public readonly uint TotalMeasures;
+        public readonly uint TicksPerMeasure;
+        public readonly double SecondsPerMeasure;
+
+        public CountdownTimeSig(uint tick, double time, uint totalMeasures, uint ticksPerMeasure, double secondsPerMeasure)
+        {
+            Tick = tick;
+            Time = time;
+            TotalMeasures = totalMeasures;
+            TicksPerMeasure = ticksPerMeasure;
+            SecondsPerMeasure = secondsPerMeasure;
+        }
+    }
+}
--- a/YARG.Core/Chart/Events/WaitCountdown.cs
+++ b/YARG.Core/Chart/Events/WaitCountdown.cs
@@ -18,8 +18,6 @@ namespace YARG.Core.Chart
 
         private List<Beatline> _measureBeatlines;
 
-        private int _measuresLeft;
-
         public WaitCountdown(List<Beatline> measureBeatlines)
         {
             _measureBeatlines = measureBeatlines;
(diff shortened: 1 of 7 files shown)


YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -278,6 +284,53 @@ protected void UpdateTimeVariables(double time)
...
+                        // Delay check until next measure if there is an active sustain
+                        if (Notes[nextNoteIndex - 1].TickEnd > State.CurrentTick)
+                        {
+                            skipCountdownCheck = true;
+                        }
+                    }
+
+                    if (!skipCountdownCheck)
+                    {
+                        // Countdown displays when the next note is farther MIN_COUNTDOWN_MEASURES
```
RileyTheFox, line 319: I think the system would work better if it required a minimum amount of time to display a countdown (for example 10 seconds), but then snapped to the next measure to base the countdown on. If you have a high bpm, 4 measures is not a long time and countdowns will appear very often. Also time signatures such as 1/x would produce countdowns in pretty much any brief break in notes.

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -278,6 +284,53 @@ protected void UpdateTimeVariables(double time)
                 State.TicksEveryBeat = nextTimeSigTick - currentBeatTick;
                 YargLogger.Assert(State.TicksEveryBeat != 0, "Ticks per beat is 0! Star Power will be NaN after this");
             }
+
+            uint lastMeasureCount = currentTimeSig.GetMeasureCount(State.LastTick, SyncTrack);
+
+            if (
```
RileyTheFox, line 290: I saw you mentioned in the Discord that it may be better to "parse" countdowns ahead of time instead of calculating when they'll be at runtime and I do think that will be better. You could create a `Countdown` object which stores the start and end time (plus any tick or measure information if needed), store a list of them and then just check for when they start and end.

Refer to `GetSoloSections` in BaseEngine.Generic.cs for how that could possibly be done

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -48,16 +50,20 @@ public abstract class BaseEngine<TNoteType, TEngineParams, TEngineStats, TEngine
 
         public delegate void SoloEndEvent(SoloSection soloSection);
 
+        public delegate void CountdownChangeEvent(uint measuresLeft);
+
         public NoteHitEvent?    OnNoteHit;
         public NoteMissedEvent? OnNoteMissed;
 
-        public StarPowerPhraseHitEvent?  OnStarPowerPhraseHit;
+         public StarPowerPhraseHitEvent?  OnStarPowerPhraseHit;
```
RileyTheFox, line 58: Some weird spacing has been added here

YARG.Core/Engine/Drums/DrumsEngine.cs
```diff
@@ -246,7 +252,7 @@ protected static int ConvertInputToPad(DrumsEngineParameters.DrumMode mode, Drum
                 {
                     DrumsAction.Kick => (int) FourLaneDrumPad.Kick,
 
-                    DrumsAction.RedDrum    => (int) FourLaneDrumPad.RedDrum,
+                     DrumsAction.RedDrum    => (int) FourLaneDrumPad.RedDrum,
```
RileyTheFox, line 255: Some more weird spacing

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -135,6 +135,12 @@ protected virtual void Overstrum()
                 return;
             }
 
+            // Cancel overstrum if WaitCountdown is active
+            if (State.CountdownMeasuresLeft > 1)
```
RileyTheFox, line 139: This should be alright, but if you can please test if the countdown end times are deterministic so that the same overstrums are always cancelled in both real run and replay runs.

RileyTheFox (changes requested):
Requesting changes mostly based on some formatting mistakes but the other comments are there more for discussion rather than necessary changes.

Got quite a lot of engine changes planned very soon which would interfere with how this works slightly so want to discuss the implementations more.

YARG.Core/Engine/BaseEngine.Generic.cs
Purplo-cf: Okay, how does this sound?

- [ ] Replace BaseEngineState.CountdownMeasuresLeft with ActiveCountdownEndTick
- [ ] Create new ChartEvent subclass Countdown and store start and length in seconds and ticks
- [ ] Create a protected Countdowns List property on BaseEngine that stores Countdowns
- [ ] Create private GetCountdowns() method that runs after GetSoloSections() that populates the list. Compare note spacing and create a Countdown instance if any notes are further than 10 seconds apart

    - Creating countdowns based off of absolute time instead of the current time signature should address your concern below. It will also display countdown numbers <4 in low BPM songs, but that may not be an issue.
    - Looping through the entire chart twice for each player is a little clunky, but there could be room here to add a OnChartAvailable callback once the chart data is converted for the current engine

- [ ] Change BaseEngine.Generic.UpdateTimeVariables() logic
   1. If Countdowns list is empty, move on
   2. If Countdowns list is populated and State.ActiveCountdownEndTick is null, compare the current tick against the start ticks and durations of the stored Countdowns
   3. If this is the first tick that enters one of the stored Countdowns, calculate length in measures based off of the current State.TicksEveryMeasure, round down so that the Countdown ends at the start of the first measure that contains a note
   4. Set State.ActiveCountdownEndTick to the rounded value **end**
   5. If State.ActiveCountdownEndTick is not null and measureCount > lastMeasureCount, calculate measuresLeft from the current tick to the stored value, and invoke OnCountdownChange(measuresLeft) as it currently exists

- [ ] Change Engine prevent overstrum/overhit logic so that it is only prevented if State.CurrentTick < State.ActiveCountdownEndTick

RileyTheFox: > Creating countdowns based off of absolute time instead of the current time signature should address your concern below. It will also display countdown numbers <4 in low BPM songs, but that may not be an issue.

You could add a minimum measure count as well but the algorithm which decides what makes a countdown can be finalised another time.

> Looping through the entire chart twice for each player is a little clunky, but there could be room here to add a OnChartAvailable callback once the chart data is converted for the current engine

Looping the data twice is fine. It happens once when the engine is initialized and having callbacks just adds unneeded complexity.

> Change Engine prevent overstrum/overhit logic so that it is only prevented if State.CurrentTick < State.ActiveCountdownEndTick

Generally any hit logic should be using time as a comparison, not ticks (because time is a continuous value while ticks are discrete values). I would also opt for a boolean instead, such as `State.IsCountdownActive` which is enabled whenever a countdown is currently present. This way the condition is simple: If a countdown is active, prevent the overstrum. No need to check ticks or times

YARG.Core/Engine/Guitar/GuitarEngine.cs
Purplo-cf: Is the below test process sufficient (after implementing the changes above) ?

- Set up a dummy chart with guitar, bass, and drum notes at the beginning and end, with a blank section in the middle
- Run all 3 instruments through the chart twice
    1. One run where an overstrum in the very beginning, middle, and end of the countdown still maintains the combo
    2. One run where an overstrum right before and right after the countdown breaks the combo

Also, do you have a preferred format for reviewing the results, or is it more of an honors system kind of deal?

RileyTheFox: > Set up a dummy chart with guitar, bass, and drum notes at the beginning and end, with a blank section in the middle

Yeah this is fine. I usually make test charts to simplify testing specific scenarios.
> * Run all 3 instruments through the chart twice
>   
>   1. One run where an overstrum in the very beginning, middle, and end of the countdown still maintains the combo
>   2. One run where an overstrum right before and right after the countdown breaks the combo

Just do 1 run where you do both at the same time. I would honestly just bind strum to scroll wheel or something and spam inputs so you get as many inputs close to the countdown times to test with as possible.

> Watch the replay data and confirm the same behavior and score

I almost never watch the replay itself. What I do is utilize our logging system and spam logs to the console, then compare the results from the "real run" and the replay simulation using the History menu's "Analyze" button (unique to the `engine-refactor` branch)

Personally what I would do is output a log when the overstrum gets cancelled by the countdown (make it say something like "`Overstrum cancelled by countdown at {time}`"), and then a log for when an overstrum is allowed through ("`Overstrummed at {time}`" for example). I copy the console output from the real run and save it somewhere. Then I clear my console and then analyze the replay and copy its console output. I filter out any garbage (like the script link formatting) then use [WinMerge](https://winmerge.org/?lang=en) to compare the 2 outputs for any differences.

You can see the engine code is littered with Trace logs throughout for testing and debugging. These are filtered out by default because the minimum log level in YARG is Debug when in editor. You can use the Debug or Trace level if you like, but you'll have to filter out a lot more logs if you use Trace so when I need to be really specific I just use Debug (or create a new temporary log level lol)

> Also, do you have a preferred format for reviewing the results, or is it more of an honors system kind of deal?

Not really. Just evidence that events from the real run and the replay have identical timings is enough because that's all that really matters.

YARG.Core/Engine/BaseEngine.Generic.cs
Purplo-cf: Okay, everything discussed above should be implemented now.
After changes to the engine this week, a lot has been moved around and most of the countdown window detection is happening in GenerateQueuedUpdates() instead of UpdateTimeVariables() now.

The new GetWaitCountdowns method is properly accounting for changes in time signature mid-chart, which the last version (and the last me) didn't think about at all.

I'm still making assumptions about one consistent tempo per time signature that it picks up, but I this seems like a good point to check in before I make this even more convoluted...

YARG.Core/Engine/Guitar/GuitarEngine.cs
Purplo-cf: Dummy chart test was a success

RileyTheFox:
Hope you don't mind the wait on this PR. Been extremely busy with lots of changes and refactors myself so haven't had the chance to look at it fully and test it.

From a glance everything looks fine but I just need to put it through some tests and that eats up a lot of time which I need to put into other things at the moment. So long as this PR doesn't get riddled with merge conflicts down the line it should be pretty much ready for me to merge it in when I get the chance.

If I come across anything that needs changing when I get around to testing it I'll report back.

Purplo-cf:
No worries! I'm learning a lot about how everything fits together. I've had time to fuss with this since the original PR and it's in a much better state now

RileyTheFox:
@Purplo-cf try not to merge commits from other branches in because it includes those changes as part of the PR. Rebase and force-push if necessary instead.

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -138,6 +175,40 @@ protected override void GenerateQueuedUpdates(double nextTime)
...
+
+            // Only check for WaitCountdowns in this chart if there are any remaining
+            if (WaitCountdowns.Count > State.CurrentWaitCountdownIndex)
+            {
+                if (!State.IsWaitCountdownActive && nextTime > WaitCountdowns[0].Time)
+                {
+                    State.CurrentWaitCountdownIndex = WaitCountdowns.GetIndexOfPrevious(nextTime);
+
+                    // A new countdown will start before nextTime
+                    State.IsWaitCountdownActive = IsTimeBetween(WaitCountdowns[State.CurrentWaitCountdownIndex].Time, previousTime, nextTime);
```
RileyTheFox, line 187: The countdown should not be started like this. The countdown's time being in between the previous frame and the next frame does NOT guarantee that the next engine update is the countdown's time. 

If an update is queued before it (such as strum leniency timer) and both are between the previous and next frame, then the countdown will be started before its actual time, which could lead to issues.

Engine state changes should never be made in the generation of queued updates. All this function should do is queue times which the engine will be ran at in between the previous and next frame. Any state changes MUST occur within the Engine update loop somewhere, which starts from [here](https://github.com/YARC-Official/YARG.Core/blob/0addf7f4b4275d1fecd3a72885a377d19209d602/YARG.Core/Engine/BaseEngine.cs#L381).

YARG.Core/Engine/Drums/DrumsEngine.cs
```diff
@@ -40,6 +40,13 @@ public virtual void Overhit()
                 return;
             }
 
+            // Cancel overhit if WaitCountdown is active
+            if (State.IsWaitCountdownActive)
+            {
+                YargLogger.LogFormatTrace(this.GetType().Name+" overhit prevented during WaitCountdown at time: {0}, tick: {1}", State.CurrentTime, State.CurrentTick);
```
RileyTheFox, line 46: Concatenating the type name results in a string allocation, even if the log level is above Trace. Put it as a parameter, but realistically I wouldn't log the type at all as it's not necessary.

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -143,6 +143,13 @@ protected virtual void Overstrum()
                 return;
             }
 
+            // Cancel overstrum if WaitCountdown is active
+            if (State.IsWaitCountdownActive)
+            {
+                YargLogger.LogFormatTrace(this.GetType().Name+" overstrum prevented during WaitCountdown at time: {0}, tick: {1}", State.CurrentTime, State.CurrentTick);
```
RileyTheFox, line 149: Same comment as put in DrumsEngine

RileyTheFox (changes requested):

YARG.Core/Engine/BaseEngine.Generic.cs
Purplo-cf: Oh weird. The first version of this _was_ started from `UpdateTimeVariables` but I had to move it to get it working after pulling the latest `engine-refactor` changes like a month ago. I put it back and it's working now, so that makes this much simpler!

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -157,40 +157,6 @@ protected override void GenerateQueuedUpdates(double nextTime)
                     QueueUpdateTime(noteBackEndIncrement, "Note Back End");
                 }
             }
-
```
RileyTheFox: You still need to queue the update for the countdown to start. The whole point of this part is you're *queuing* the update to happen for future execution. You don't want to directly modify countdown status here, you just want to queue the **time** when it should happen. The state change then happens in UpdateTimeVariables when it is invoked on the countdown's exact time.

So here you still need to check if the next countdown to happen is between the previous and next time, if so, queue its time and then have the UpdateTimeVariables portion handle the enabling/disabling of it. Make sure to queue the start and end time appropriately.

If you're not sure what I mean by this, look at any part of the GuitarEngine code in `GenerateQueuedUpdates` and look what it does. None of it modifies state, just checks timers or note times and queues updates as appropriate. Then when the engine is executed at that update's time, the state change will occur

RileyTheFox (changes requested):

YARG.Core/Engine/BaseEngine.Generic.cs
Purplo-cf: Thanks, I put the queued updates back in, and added a couple of log statements to compare the times that the countdowns are turned on and off, compared to the expected start and end times. Here's what I got on the guitar part for Won't Get Fooled Again
image

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -138,6 +157,28 @@ protected override void GenerateQueuedUpdates(double nextTime)
...
+            if (State.CurrentWaitCountdownIndex < WaitCountdowns.Count)
+            {
+                // Queue updates for countdown start/end/change
+                var currentCountdown = WaitCountdowns[State.CurrentWaitCountdownIndex];
+                
+                if (State.IsWaitCountdownActive)
+                {
+                    double changeTime = currentCountdown.GetNextUpdateTime();
+
+                    if (IsTimeBetween(changeTime, previousTime, nextTime))
```
RileyTheFox, line 170: Tested this PR and discovered that this condition is not passing in runs except for the start and end times. All the measure times in between are not passing this, causing the update times to be inconsistent.

Comparatively in replays this condition does pass, but only for odd amounts of measures.

Log dump of real run: https://pastebin.com/yWRdRr4B
Log dump of replay: https://pastebin.com/kFasQwjy

I've attached a patch with additional logging to help you debug this
[countdown logging.patch](https://github.com/user-attachments/files/15973820/countdown.logging.patch)

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -158,6 +199,44 @@ protected override void UpdateTimeVariables(double time)
...
+                    }
+                }
+
+                if (State.IsWaitCountdownActive)
+                {
+                    var activeCountdown = WaitCountdowns[State.CurrentWaitCountdownIndex];
+
+                    int countdownMeasuresRemaining = activeCountdown.GetRemainingMeasures(State.CurrentTick);
+
+                    UpdateCountdown(countdownMeasuresRemaining);
```
RileyTheFox, line 230: This is not being called at consistent times for any measures in between the start and end. This is ok for the intermittent times in between measures, but when the `countdownMeasuresRemaining` value changes, the time of this should ideally be getting queued but as mentioned in the previous comment, something is preventing that.

RileyTheFox (changes requested):

Purplo-cf:
`GetRemainingMeasures` should be sending out the correct queue times now. I'm also throttling `UpdateCountdown` now so that it only fires when the number of remaining measures has changed. The logs should be much easier to read now:

Actual run:
https://pastebin.com/CWHtRAP6

Replay of same run:
https://pastebin.com/kpeT8HkF

YARG.Core/Chart/Events/WaitCountdown.cs
```diff
@@ -39,25 +39,33 @@ public WaitCountdown(List<Beatline> measureBeatlines)
             _measuresLeft = TotalMeasures;
         }
 
-        public int GetRemainingMeasures(uint currentTick)
+        public int GetRemainingMeasures(uint currentTick, out bool needsUpdate)
```
RileyTheFox, line 42: I don't think this is the best way of handling checking for a change in measures. I would opt for making `_measuresLeft` into a public get private setter property that the BaseEngine can access. This function should only handle returning however many measures are left, not signalling if something should be updated.

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -225,15 +225,18 @@ protected override void UpdateTimeVariables(double time)
                 {
                     var activeCountdown = WaitCountdowns[State.CurrentWaitCountdownIndex];
 
-                    int countdownMeasuresRemaining = activeCountdown.GetRemainingMeasures(State.CurrentTick);
+                    int countdownMeasuresRemaining = activeCountdown.GetRemainingMeasures(State.CurrentTick, out bool needsUpdate);
 
-                    UpdateCountdown(countdownMeasuresRemaining);
-
-                    if (countdownMeasuresRemaining <= WaitCountdown.END_COUNTDOWN_MEASURE)
+                    if (needsUpdate)
```
RileyTheFox: Instead of using a bool from the function call above, just store the remaining amount of measures before calling the function (the public property) into a local variable, then call the function and compare the 2 values for the condition

YARG.Core/Chart/Events/WaitCountdown.cs
```diff
@@ -5,7 +5,7 @@ namespace YARG.Core.Chart
 {
     public class WaitCountdown : ChartEvent
     {
-        public const float MIN_SECONDS = 10;
+        public const float MIN_SECONDS = 9;
```
RileyTheFox, line 8: What was the need for this change?

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -216,6 +217,7 @@ protected override void UpdateTimeVariables(double time)
                         {
                             // Entered new countdown window
                             State.IsWaitCountdownActive = true;
+                            firstCountdownFrame = true;
```
RileyTheFox, line 220: Instead of having this flag to work around an issue, I'd prefer to just call UpdateCountdown() here instead.

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -227,7 +229,7 @@ protected override void UpdateTimeVariables(double time)
 
                     int countdownMeasuresRemaining = activeCountdown.GetRemainingMeasures(State.CurrentTick, out bool needsUpdate);
```
RileyTheFox, line 230: With the previous comment, change this part into an `else` statement. This will prevent the countdown updating part from executing if a countdown was not active and has just been started in this call.

RileyTheFox (changes requested):

YARG.Core/Chart/Events/WaitCountdown.cs
Purplo-cf: Just calibrating. There was a 7 measure wait in my test chart (3/4@120bpm) that was getting skipped because it was just under the threshold, but still felt long enough to show a countdown. The original 10 second threshold was an arbitrary starting point.

YARG.Core/Engine/BaseEngine.Generic.cs
Purplo-cf: No longer necessary. The new logic for detecting a change in `MeasuresRemaining` gets around the original issue that this was trying to fix.

RileyTheFox (changes requested):
Discovered some more inconsistent update times through some testing:

First attempt:
image

Second attempt:
image

Blank lines added where logs are missing just so the other entries line up.

I think I know what the culprit is: Large gaps in-between input times results in some of the update times for countdowns getting skipped. The first time I came across it I was confused but had a theory, then I tested that theory and it was confirmed. In the second attempt I waited around 5 seconds into the song before pressing anything and there was a few missing updates.

Song used for testing was Soulless 5 as it has an 18 second countdown right at the start so is easy to test with.

RileyTheFox:
Seems to have been a bug in the general update loop for the engine. Will keep testing things however

RileyTheFox (approved):

PR #174, opened 2024-05-29
Change points awarded per vocals phrase for lower difficulties

poliwhirlge:
Up to now we've been awarding 2000 base points per completed phrase regardless of difficulty, this PR scales the values for lower difficulties similarly to how RB's scoring works.

### DIFF
--- a/YARG.Core/Engine/Vocals/VocalsEngine.cs
+++ b/YARG.Core/Engine/Vocals/VocalsEngine.cs
@@ -171,9 +171,15 @@ namespace YARG.Core.Engine.Vocals
             }
         }
 
+        /// <remarks>
+        /// Because this method is called in the constructor of
+        /// <see cref="BaseEngine{TNoteType,TEngineParams,TEngineStats,TEngineState}"/>,
+        /// it must use the value in EngineParameters instead of the property
+        /// POINTS_PER_PHRASE as it is not yet set.
+        /// </remarks>
         protected sealed override int CalculateBaseScore()
         {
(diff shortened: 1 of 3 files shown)


YARG.Core/Engine/Vocals/VocalsEngineParameters.cs
```diff
@@ -19,17 +19,20 @@ public class VocalsEngineParameters : BaseEngineParameters
         /// </summary>
         public bool SingToActivateStarPower { get; private set; }
 
+        public int PointsPerPhrase { get; private set; }
```
EliteAsian123, line 22: Make sure the engine param is added to the serialization methods!

YARG.Core/Engine/Vocals/VocalsEngine.cs
```diff
@@ -7,7 +7,7 @@ namespace YARG.Core.Engine.Vocals
     public abstract class VocalsEngine :
         BaseEngine<VocalNote, VocalsEngineParameters, VocalsStats, VocalsEngineState>
     {
-        protected const int POINTS_PER_PHRASE = 2000;
+        protected int POINTS_PER_PHRASE;
```
EliteAsian123, line 10: It may be best to just completely remove this field and access the amount of points from the engine params

EliteAsian123 (changes requested):

YARG.Core/Engine/Vocals/VocalsEngineParameters.cs
poliwhirlge: Yeah I thought that this was for user configured presets, but it's for replays, my bad.

PR #178, opened 2024-06-07
Drum velocity core

Purplo-cf:
Hit logic for accepting velocity-based inputs on DrumNotes.

Added properties to DrumNote, DrumsEngine, and DrumEngineState for storing velocity data on hit.

Updated MidReader to accept "ENABLE_CHART_DYNAMICS" text event without brackets.


YARG.Core/Chart/Notes/DrumNote.cs
Purplo-cf, line 1: Velocity bonus points are awarded based on the last "pressed" amount of this input compared to the VELOCITY_THRESHOLD constant.

Default: Ghosts are <= VELOCITY_THRESHOLD, Accents are >= (1 - VELOCITY_THRESHOLD).

Context sensitive: A note on this pad was hit within the last 3 seconds. Ghosts notes will accept (lastVelocity - VELOCITY_THRESHOLD) if it is higher than the default VELOCITY_THRESHOLD, and vice versa for Accent notes

YARG.Core/Engine/Drums/DrumsEngineState.cs
Purplo-cf, line 1: Storing the last input axis value in DrumsEngineState as HitVelocity. Gets sent to DrumNote.SetHitState()

YARG.Core/Engine/Drums/DrumsEngine.cs
Purplo-cf, line 1: Notes are awarded a velocity bonus of (POINTS_PER_NOTE * 0.5), when velocity conditions are met. This is currently a bonus of +25 for a total of 75 points in Standard Drums or 85 points in Pro Drums.

This logic is consistent with how GHWT handled accent bonuses, but I couldn't find any actual numbers for what Clone Hero does here

YARG.Core/Chart/Notes/DrumNote.cs
```diff
@@ -1,9 +1,16 @@
...
+using YARG.Core.Logging;
 
 namespace YARG.Core.Chart
 {
     public class DrumNote : Note<DrumNote>
     {
+        private const float VELOCITY_THRESHOLD = 0.35f;
+        
+        // The maximum allowed time (seconds) between notes to use context-sensitive velocity scoring
+        private const float SITUATIONAL_VELOCITY_WINDOW = 3;
```
EliteAsian123, line 13: These should be moved to the engine or even be made into engine parameters (see comments below on why)

YARG.Core/Chart/Notes/DrumNote.cs
```diff
@@ -15,6 +22,9 @@ public class DrumNote : Note<DrumNote>
         public bool IsAccent  => Type == DrumNoteType.Accent;
         public bool IsGhost   => Type == DrumNoteType.Ghost;
 
+        public float HitVelocity = -1;
```
EliteAsian123, line 25: Instead of using `-1` as a "not set" value, use a nullable float (`float?`) instead, which makes it explicit that `null` is not set.

YARG.Core/Chart/Notes/DrumNote.cs
```diff
@@ -50,6 +60,74 @@ public override void ResetNoteState()
...
+                    YargLogger.LogFormatDebug("Ghost note was hit with a velocity of {0} at tick {1}. Bonus awarded: {2}", velocity, Tick, AwardVelocityBonus);
+                }
+                else if (IsAccent)
+                {
+                    AwardVelocityBonus = velocity > (1 - awardThreshold);
+                    YargLogger.LogFormatDebug("Accent note was hit with a velocity of {0} at tick {1}. Bonus awarded: {2}", velocity, Tick, AwardVelocityBonus);
+                }
+            }
+            
+            SetHitState(hit, includeChildren);
```
EliteAsian123, line 130: This code for checking previous notes should be moved to the YARG engine file as how these things are calculated can be changed based on what engine is being used. The note class should only contain the data and methods to manipulate the data instead of actual logic.

YARG.Core/Engine/Drums/DrumsEngine.cs
```diff
@@ -167,6 +167,12 @@ protected int GetPointsPerNote()
         protected override void AddScore(DrumNote note)
         {
             int pointsPerNote = GetPointsPerNote();
+
+            if (note.AwardVelocityBonus)
+            {
+                pointsPerNote += (int)(POINTS_PER_NOTE * 0.5);
+            }
+
```
EliteAsian123, line 175: Once the velocity check is moved into the YARG engine class the `AwardVelocityBonus` check shouldn't be needed and the bonus can just be directly added to the score.

YARG.Core/Engine/Drums/DrumsEngineState.cs
```diff
@@ -9,12 +9,14 @@ public class DrumsEngineState : BaseEngineState
         /// be based off of <see cref="FourLaneDrumPad"/> or <see cref="FiveLaneDrumPad"/>.
         /// </summary>
         public int? PadHit;
+        public float HitVelocity = 1;
```
EliteAsian123, line 12: Like in the note class, this should be a nullable float, much like the nullable int above.

EliteAsian123 (changes requested):
Hi, thanks for the PR, please see the requested changes below, and on the YARG-side PR.

Thanks!

YARG.Core/Engine/Drums/DrumsEngine.cs
```diff
@@ -8,9 +8,14 @@ namespace YARG.Core.Engine.Drums
     public abstract class DrumsEngine : BaseEngine<DrumNote, DrumsEngineParameters,
         DrumsStats, DrumsEngineState>
     {
+        private const float VELOCITY_THRESHOLD = 0.35f;
+        
+        // The maximum allowed time (seconds) between notes to use context-sensitive velocity scoring
+        private const float SITUATIONAL_VELOCITY_WINDOW = 1.5f;
```
EliteAsian123, line 14: These should both be made into engine parameters so they can be customized with engine presets.

YARG.Core/Engine/Drums/DrumsEngine.cs
```diff
@@ -125,6 +131,77 @@ protected void HitNote(DrumNote note, bool activationAutoHit)
...
+                    awardVelocityBonus = lastInputVelocity < awardThreshold;
+                    YargLogger.LogFormatTrace("Ghost note was hit with a velocity of {0} at tick {1}. Bonus awarded: {2}", lastInputVelocity, hitNote.Tick, awardVelocityBonus);
+                }
+                else if (hitNote.IsAccent)
+                {
+                    awardVelocityBonus = lastInputVelocity > (1 - awardThreshold);
+                    YargLogger.LogFormatTrace("Accent note was hit with a velocity of {0} at tick {1}. Bonus awarded: {2}", lastInputVelocity, hitNote.Tick, awardVelocityBonus);
+                }
+
+                hitNote.AwardVelocityBonus = awardVelocityBonus;
```
EliteAsian123, line 201: Instead of using a variable stored within the note class here, the score can probably just be manipulated directly. This eliminates the need for an extra bool within the note class.

EliteAsian123 (changes requested):
Hello!

It appears two of my suggestions have not been taken into account so I've requested changes again below.

If there is a reason you're not doing this any of these, please let me know!

YARG.Core/Engine/Drums/DrumsEngine.cs
Purplo-cf: Oh duh, EngineParameters, sorry! I thought you just meant to move the constants into the DrumsEngine file lol

YARG.Core/Engine/Drums/DrumsEngine.cs
Purplo-cf: I wanted to try a Deadly Dynamics modifier after this and wasn't sure how else to do it. The score is currently being added after the hit state is set to true. With the modifier, we would have to check the velocity before setting the hit state in case we need to count it as a miss instead.

With that and the bonus points check I'm storing the pass/fail result in the note so we only have to evaluate that once.

EliteAsian123: Yes so instead, what can happen, is that this is calculated right above line 73 here and the note can be denied the hit in the YARG drums engine. As for the score, it can be added after the note is hit, or can be passed into the `HitNote` method somehow.

https://github.com/YARC-Official/YARG.Core/blob/bb8829a0d8620f26f07ebc799a0248dbad08404d/YARG.Core/Engine/Drums/Engines/YargDrumsEngine.cs#L70-L79

YARG.Core/Engine/Drums/DrumsEngineParameters.cs
```diff
@@ -16,6 +16,12 @@ public enum DrumMode : byte
         /// </summary>
         public DrumMode Mode { get; private set; }
 
+        //Ghost notes are below this threshold, Accent notes are above 1 - threshold
+        public float VelocityThreshold = 0.35f;
+        
+        // The maximum allowed time (seconds) between notes to use context-sensitive velocity scoring
+        public float SituationalVelocityWindow = 1.5f;
+
```
EliteAsian123, line 24: These should be properties (`{ get; private set; }`) and be set through the constructor. Also, make sure to serialize these new properties!

[Here](https://github.com/YARC-Official/YARG.Core/blob/engine-refactor/YARG.Core/Engine/Guitar/GuitarEngineParameters.cs) is an example from the guitar params.

YARG.Core/Engine/Drums/DrumsEngine.cs
```diff
@@ -7,10 +7,10 @@ namespace YARG.Core.Engine.Drums
 {
     public abstract class DrumsEngine : BaseEngine<DrumNote, DrumsEngineParameters,
         DrumsStats, DrumsEngineState>
-    {
```
EliteAsian123, line 10: Random space appeared here

EliteAsian123 (changes requested):
A few things about the engine parameters. Be sure to check my comment on the non-resolved issues above!

YARG.Core/Engine/Drums/DrumsEngine.cs
Purplo-cf: Thanks, I put the logic where you suggested and I'm no longer storing `ApplyVelocityBonus` in each individual note. The `DrumsEngine.ApplyVelocity` method now just returns a bool on whether a hit qualifies for bonus points. That bool goes into a new `HitNote` overload in `YargDrumsEngine`. I left a TODO in there on where I think we can capture that bool and go to `MissNote` instead if Deadly Dynamics is on.

YARG.Core/Engine/Drums/Engines/YargDrumsEngine.cs
```diff
@@ -93,7 +98,18 @@ protected override void CheckForNoteHit()
             if (State.PadHit != null)
             {
                 Overhit();
-                State.PadHit = null;
+                ResetPadState();
+            }
+        }
+
+        protected void HitNote(DrumNote note, bool awardVelocityBonus, bool activationAutoHit)
```
EliteAsian123, line 105: Creating an overload here isn't the clearest. It's probably better to call this something else or to just put it above with the call to the original `HitNote`.

EliteAsian123 (changes requested):
One small nitpick and this should be good to go! I'm not going to worry about the art on the YARG-side PR and I'll implement the UI part of the engine parameters for you. Thanks!

YARG.Core/Engine/Drums/Engines/YargDrumsEngine.cs
Purplo-cf: I moved it up to the original `HitNote` like you suggested, since it's only being used once

Purplo-cf:
> I'll implement the UI part of the engine parameters for you.

Thanks! For future reference, what other files should I have updated?

EliteAsian123:
> Thanks! For future reference, what other files should I have updated?

It's quite simple, keep a look out for the commits I'll make which will address this after it's merge!

PR #182, opened 2024-06-13
Utilize A Hierarchy Of FixedArray Types To Enhance Memory Usage (Part 1)

sonicfind:
The main aim of these next few sequences of PRs is to tackle some memory usage and GC issues I've observed during scanning. Note: any speed improvements are purely consequential.

To add MemoryMappedFile functionality, it was necessary to split the FixedArray type into pieces. At base, the FixedArray type is just an alias. It's sole function is to hold the pointer and length for a block of memory. The subtypes would handle any allocation or mapping of the block of memory to be represented.

Usage of memory mapped files necessitated holding onto the length of the file - as the MemoryMappedFile might not provide the exactly correct value (this was tested). As such, this also brings along a AbridgedFileInfo_Length struct to coincide with the original.

Through these array types, we can improve parsing strategies for scanning. As the memory for files remains in-place, we have access to pointer logic - which would provide better cache locality and lead to better performance (although, most of that will be in Part 2)

### DIFF
--- a/YARG.Core/IO/AbridgedFileInfo.cs
+++ b/YARG.Core/IO/AbridgedFileInfo.cs
@@ -4,83 +4,74 @@ using YARG.Core.Extensions;
 
 namespace YARG.Core.IO
 {
+    public interface IAbridgedInfo
+    {
+        public string FullName { get; }
+        public DateTime LastUpdatedTime { get; }
+    }
+
     /// <summary>
     /// A FileInfo structure that only contains the filename and time last added
     /// </summary>
-    public sealed class AbridgedFileInfo
+    public readonly struct AbridgedFileInfo : IAbridgedInfo
     {
         public const FileAttributes RECALL_ON_DATA_ACCESS = (FileAttributes)0x00400000;
 
         /// <summary>
         /// The flie path
         /// </summary>
-        public readonly string FullName;
+        public readonly string _fullname;
 
         /// <summary>
         /// The time the file was last written or created on OS - whichever came later
         /// </summary>
-        public readonly DateTime LastUpdatedTime;
+        public readonly DateTime _lastUpdatedTime;
 
-        /// <summary>
-        /// The length of the file in bytes
-        /// </summary>
-        public readonly long Length;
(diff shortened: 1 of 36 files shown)


YARG.Core/IO/AbridgedFileInfo.cs
```diff
@@ -4,33 +4,42 @@
...
     /// <summary>
     /// A FileInfo structure that only contains the filename and time last added
     /// </summary>
-    public sealed class AbridgedFileInfo
+    public readonly struct AbridgedFileInfo : IAbridgedInfo
     {
         public const FileAttributes RECALL_ON_DATA_ACCESS = (FileAttributes)0x00400000;
 
         /// <summary>
         /// The flie path
```
EliteAsian123, line 21: flie path

YARG.Core/IO/AbridgedFileInfo.cs
```diff
@@ -4,33 +4,42 @@
...
         /// </summary>
-        public readonly string FullName;
+        public readonly string _fullname;
 
         /// <summary>
         /// The time the file was last written or created on OS - whichever came later
         /// </summary>
-        public readonly DateTime LastUpdatedTime;
+        public readonly DateTime _lastUpdatedTime;
```
EliteAsian123, line 28: `_fullname` along with `_lastUpdatedTime` can be made private (which I'm assuming is the intention as the naming scheme was changed, could be wrong though if there are other changes).

EliteAsian123 (changes requested):
Just a couple of style related things I noticed.

I played a few songs and scanned my library a few times and didn't notice any issues, so pretty much approved by be👍

YARG.Core/IO/AbridgedFileInfo.cs
sonicfind: Ah right forgot about those

YARG.Core/IO/AbridgedFileInfo.cs
sonicfind: LMAO

PR #186, opened 2024-06-15
Drum Hit SFX Core

raphaelgoulart:
Companion to https://github.com/YARC-Official/YARG/pull/802; good to be reviewed though


raphaelgoulart:
Actually marking this as ready to review since it doesn't impact YARG itself (and even fixes a bug with 127-velocity hits not being detected, though maybe not in the best of ways)

YARG.Core/Engine/Drums/Engines/YargDrumsEngine.cs
```diff
@@ -15,8 +15,9 @@ public class YargDrumsEngine : DrumsEngine
 
         protected override void MutateStateWithInput(GameInput gameInput)
         {
-            if (gameInput.Button)
+            if (gameInput.Axis > 0) // For some reason, it's ignoring 127-velocity hits and this fixes it...
```
RileyTheFox, line 18: Would suggest removing this comment as we discussed why this was the case on Discord. Perhaps replace it with a comment detailing why Axis is used here instead of Button (in regards to the floating point bit values)

Also try to place comments a line above, instead of on the same line as an expression or statement

raphaelgoulart: Done, sadly I couldn't explain it in briefer terms but hopefully my explanation works. Let me know if there's a better way to clarify it

YARG.Core/Audio/DrumSampleChannel.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public void Play(double volume)
+        {
+            lock (this)
+            {
+                if (!_disposed)
+                {
+                    SetVolume_Internal(volume);
+                    Play_Internal();
+                }
```
sonicfind, line 31: Set the `_volume` member variable to the initial value of volume passed to this function. Afterwards,  multiply that by the current `DrumSfx` `StemSettings` value before sending it `SetVolume_Internal()`.

YARG.Core/Audio/DrumSampleChannel.cs
```diff
@@ -0,0 +1,79 @@
...
+        private void SetVolume(double volume)
+        {
+            lock (this)
+            {
+                if (!_disposed)
+                {
+                    SetVolume_Internal(volume);
+                }
+            }
+        }
```
sonicfind, line 44: When `StemSettings.OnVolumeChange` calls this function, multiply the value of `volume` by `_volume` before sending it to `SetVolume_Internal()`.

YARG.Core/Audio/DrumSampleChannel.cs
```diff
@@ -0,0 +1,79 @@
...
+        protected readonly string _path;
+        protected readonly int _playbackCount;
+
+        public readonly DrumSfxSample Sample;
+        protected DrumSampleChannel(DrumSfxSample sample, string path, int playbackCount)
+        {
+            Sample = sample;
+            _path = path;
+            _playbackCount = playbackCount;
+        }
```
sonicfind, line 21: Attach the `StemSettings.OnVolumeChange` for DrumSfx to `SetVolume()`

YARG.Core/Audio/DrumSampleChannel.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace YARG.Core.Audio
+{
+    public abstract class DrumSampleChannel : IDisposable
+    {
+        public const int ROUND_ROBIN_MAX_INDEX = 3;
+        private bool _disposed;
+
+        protected readonly string _path;
+        protected readonly int _playbackCount;
+
```
sonicfind, line 14: Add a `double _volume` member variable here.

sonicfind (changes requested):
Pretty solid: just need that `_volume` member variable to keep everything in-line.

sonicfind (approved):

PR #189, opened 2024-06-20
Generate drum fill phrases

Purplo-cf:
This is a revised version of my earlier PR based off of `master` instead of `engine-refactor`.

I've made the formatting changes that EA suggested, and all functionality has been moved to SongChart so that the chart modifications will show up in replays.

I have not made much progress in simplifying the algorithm, but I did change it so that the chart will only be parsed once for activation phrase placement instead of running on every available difficulty like before.

Ending note placement for each fill phrase is still being evaluated per-difficulty


YARG.Core/Chart/Events/Phrase.cs
Purplo-cf, line 1: I added this overload just to help with readability when the new Phrase is created at the end of `ParseForActivationPhrases`. I can remove this and replace it with the 4 arg version if you'd prefer

https://github.com/YARC-Official/YARG.Core/blob/929ffd26531ab561c81d8e31e6ff39595e8c4d82/YARG.Core/Chart/SongChart.cs#L453-L457

Purplo-cf:
In the interest of cutting the code down, would you rather get rid of any of these?

- Place activators 2 measures apart instead of 4 in cases of slow tempos or long measures
- Avoid placing an activator before sections that don't have a lot of notes
- Try to line up activators with crash/kick/snare accents that are right before or right after the bar line

EliteAsian123:
I don't know how I feel about using sections within the generated drum-fill phrases. They are completely arbitrary and *could* be ridiculously long which could mess up the generation code, or could be randomly placed which could have the same effect. Placing the activators based on tempo/measures and notes should be good enough and would avoid weird edge cases, so I think it's probably best if it's done without considering sections.

I'm not really an expert when it comes to activator generation, so I may be wrong, but that's just my first thought. 

Thanks!

YARG.Core/Chart/Events/Phrase.cs
TheNathannator: I don't think there's any point to this constructor when you're going to be setting the properties anyways. In this scenario, I wrap function/constructor arguments across multiple lines:

```cs
var newDrumFillPhrase = new Phrase(
    PhraseType.DrumFill,
    fillPhraseStartTime,
    currentMeasureTime - fillPhraseStartTime,
    fillPhraseStartTick,
    currentMeasureTick - fillPhraseStartTick
); 
```

TheNathannator:
I still need to look at the specifics of the algorithm here, but here's my comments:

I don't think section placement issues should be a notable concern, if a chart has authored sections, we can *generally* assume them to be good. If they're not, that's on the charter for providing junk in their chart. Factoring in sections will result in much better activation generation, since they provide more context about notable/important spots in the song.

Considerations for long sections or charts with no sections should definitely be made if they aren't already, the 4-8 measure rule should still be in effect. Sections would moreso be a reference point for that rule.

I'm not personally concerned with cutting the code down unnecessarily, so long as it doesn't have a noticeable load time impact and produces good enough results. If you want to organize things a little better, mark `SongChart` as `partial` and move the auto-generation code to a `SongChart.AutoGeneration.cs` file.

Purplo-cf:
I moved my stuff into a partial class file that's dedicated to augmenting the existing chart data after it's loaded. I thought the existing `PostProcessSections` and `FixDrumPhraseEnds` methods fit under that category so I brought them along too.

I'm still using section markers to reset the `every 4 measures` counter, but I've added an additional check for time signature changes to do the same if the current section has not changed.

https://github.com/YARC-Official/YARG.Core/blob/9e3cd24f82af2723f7c8e357f1c822287fe8f329/YARG.Core/Chart/SongChart.AutoGeneration.cs#L236C17-L261C18

Purplo-cf:
As far as performance goes, I plugged in a full album chart for Dream Theater's Octavarium (runtime 1hr 13 minutes). It generated 334 fill phrases in 23.1 ms

PR #192, opened 2024-07-01
Whammy Pitch Modulation

jbienz:
Added pitch bending whammy support back in. I will submit another PR for YARG front-end consuming these changes and open a discussion in this feature request:

https://github.com/YARC-Official/YARG/issues/786

Here are the changes I implemented to restore this feature:

1. Added WhammyPitch (float) to StemSettings

2. Added OnWhammyPitchChange to StemSettings

3. In StemChannel subscribed to OnWhammyPitchChange at the same location volume change is called

4. In GlobalAudioHandler added Get/SetWhammyPitchSetting 

5. In StemState added SetWhammyPitch.

6. In SettingsManager.Settings uncommented UseWhammyFx setting and UseWhammyFxChange.

7. In SettingsManager.Settings uncommented WhammyPitchShiftAmount setting and WhammyPitchShiftAmountChange.

8. In GameManager.Audio added ChangeStemWhammyPitch

9. In FiveFretPlayer added SetStemWhammyPitch

10. In FiveFretPlayer called SetStemWhammyPitch in override for OnInputQueued when WhammyFactor is updated.

11. In GlobalAudioHandler changed WHAMMY_FFT_DEFAULT to 512 to try and reduce latency between guitar track and other tracks when pitch change is enabled.

### DIFF
--- a/YARG.Core/Audio/StemSettings.cs
+++ b/YARG.Core/Audio/StemSettings.cs
@@ -8,8 +8,10 @@ namespace YARG.Core.Audio
 
         private Action<double>? _onVolumeChange;
         private Action<bool>? _onReverbChange;
+        private Action<float>? _onWhammyPitchChange;
         private double _volume;
         private bool _reverb;
+        private float _whammyPitch;
 
         public StemSettings()
         {
@@ -26,6 +28,12 @@ namespace YARG.Core.Audio
         {
             add { _onReverbChange += value; }
             remove { _onReverbChange -= value; }
+        }
+
+        public event Action<float> OnWhammyPitchChange
+        {
+            add { _onWhammyPitchChange += value; }
+            remove { _onWhammyPitchChange -= value; }
         }
 
         public double VolumeSetting
@@ -49,6 +57,20 @@ namespace YARG.Core.Audio
                 {
                     _reverb = value;
                     _onReverbChange?.Invoke(value);
+                }
+            }
+        }
+
+        public float WhammyPitch
+        {
+            get => _whammyPitch;
+            set
+            {
+                value = Math.Clamp(value, 0, 1);
+                if (value != _whammyPitch)
+                {
+                    _whammyPitch = value;
+                    _onWhammyPitchChange?.Invoke(value);
                 }
             }
         }
--- a/YARG.Core/Audio/GlobalAudioHandler.cs
+++ b/YARG.Core/Audio/GlobalAudioHandler.cs
@@ -107,6 +107,16 @@ namespace YARG.Core.Audio
             StemSettings[stem].Reverb = reverb;
         }
 
+        public static float GetWhammyPitchSetting(SongStem stem)
+        {
+            return StemSettings[stem].WhammyPitch;
+        }
+
+        public static void SetWhammyPitchSetting(SongStem stem, float percent)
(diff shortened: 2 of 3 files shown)


YARG.Core/Audio/StemSettings.cs
```diff
@@ -54,5 +62,19 @@ public bool Reverb
                 }
             }
         }
+
+        public float WhammyPitch
+        {
+            get => _whammyPitch;
+            set
+            {
+                value = Math.Clamp(value, 0, 1);
```
sonicfind, line 71: This clamp will break how whammy works. The player should be able to shift in either direction, but without allowing down to -1, you'll only get pitch shift up (or down if the value is negated)

sonicfind (changes requested):

YARG.Core/Audio/StemSettings.cs
sonicfind: Actually, to add to this: what matters more than clamping to the [-1,1] range is establishing a deadzone area. Anything within the deadzone should be 0, with everything resting outside being normalized to the [0, -1/1] range starting at the deadzone.

Without a deadzone, you run into the issue where the tiniest unintentional tilt of whammy may lead to a unwanted pitch shift. This would need to be customizable to a degree.

sonicfind (changes requested):

YARG.Core/Audio/StemSettings.cs
sonicfind: I will admit though: if pitch shift up was never a thing to begin with, then oh well.

RileyTheFox: > Without a deadzone, you run into the issue where the tiniest unintentional tilt of whammy may lead to a unwanted pitch shift. This would need to be customizable to a degree.

Aren't deadzones handled on the input side in the main repo anyway? Axis inputs do have deadzones already. not sure if you want to smooth out the jump from 0 to out of deadzone though.

TheNathannator: The axis deadzone calculation already smooths out that jump, it completely re-normalizes the value.

sonicfind: perfect then

jbienz:
Hey folks. I'm finally back from "vacation". We were one of the unlucky families that got stranded with cancelled flights but luckily we're back home now.

A few thoughts:

It appeared to me that Whammy was already clamped at 0 - 1 elsewhere. For example, in the BassAudio implementation whammy pitch is already clamped in this range:

https://github.com/YARC-Official/YARG/blob/a97ea45eb60e6355fafebc91191673df32dc82c7/Assets/Script/Audio/Bass/BassAudioManager.cs#L416

I'm happy to see pitch bending in both directions as that would be more "accurate". But I did want to point out that none of my guitars physically stop at the mid-point. They are all spring-loaded to fully return to one extent of the axis. Which would mean that these guitars would always be pushed off center of the axis by default. At minimum we would need to account for that in settings.

@sonicfind, since I don't believe this appears to have been supported previously, and since many hardware peripherals don't appear to be built to support it, can we accept this PR as-is and consider your suggestion as a recommendation for a future improvement?

PR #193, opened 2024-07-11
Implement guitar note shuffle

TheNathannator:
Works great with the songs I've tried it on so far (for the most part just Soulless 5 lol), and note distribution is pretty good (albeit chaotic). Probably should have some further testing to review how exactly the shuffle characteristics go, but I feel confident in the current algorithm as an initial implementation.


YARG.Core/Chart/Notes/VocalNote.cs
```diff
@@ -182,6 +182,12 @@ public override void AddChildNote(VocalNote note)
             TotalTickLength = _childNotes[^1].TickEnd - Tick;
         }
 
+        public override void RemoveChildNote(VocalNote note)
+        {
+            // TODO
+            throw new NotImplementedException();
+        }
+
```
TheNathannator, line 190: Wasn't sure what to do here and didn't feel like going through the `AddChildNote` override at the moment, so left it as a TODO.

YARG.Core/Game/YargProfile.cs
```diff
@@ -115,7 +116,7 @@ public void CopyModifiers(YargProfile profile)
             CurrentModifiers = profile.CurrentModifiers;
         }
 
-        public void ApplyModifiers<TNote>(InstrumentDifficulty<TNote> track) where TNote : Note<TNote>
+        public void ApplyModifiers<TNote>(SongEntry song, InstrumentDifficulty<TNote> track) where TNote : Note<TNote>
```
TheNathannator, line 119: Having to pass in the SongEntry feels a little hacky, but I'm not sure how else to get the song hash in for note shuffle.

TheNathannator (commented):
Some self-comments I have:

EliteAsian123 (approved):

RileyTheFox:
Think the algorithm needs more tuning, or just removing the weighted consideration for the time being because its far too lenient. Loaded up Knights of Cydonia and the chart is different but not really shuffled at all. Even the slow chords in the intro are keeping the same frets which is rather boring.

TheNathannator:
Might be good to separate the weighted shuffling into its own modifier entirely, and maybe up the NPS needed to achieve the maximum weight.

I'm conflicted on whether or not to allow these "invalid" HOPO scenarios. On one hand they seem like they could be interesting in some scenarios, but on the other hand they're a little hard to read (though you could say the same for the tap scenario). Wouldn't be difficult to disallow them though.

PR #194, opened 2024-07-13
Fixed drum parsing ending too early

iceypotato:
This should fix issue #734 in the YARG repository. While debugging, I found that YARG will stop parsing the chart when it finds all difficulties and would not continue parsing to check for notes that would make it a 5 lane. For GHWT charts, because expert+ did not exist, it would be able to parse the notes that would make the chart a five-lane.


sonicfind:
Update the cache version in CacheHandler.cs and we’ll be good.

sonicfind:
Actually, I'll just make a separate commit for that

PR #195, opened 2024-07-14
Countdown rework core

Purplo-cf:
Removed "Get Ready!" message and added a progress bar like Clone Hero's implementation. Only start and end times are queued, while mid-countdown updates occur during every frame that `BaseEngineState.IsCountdownActive` is set to true.

### DIFF
--- a/YARG.Core/Engine/BaseEngine.Generic.cs
+++ b/YARG.Core/Engine/BaseEngine.Generic.cs
@@ -663,7 +663,7 @@ namespace YARG.Core.Engine
                     {
                         // Skip counting on measures that are too close together
                         if (beatlinesThisCountdown.Count == 0 || 
-                            curMeasureline.Time - beatlinesThisCountdown.Last().Time >= WaitCountdown.MIN_UPDATE_SECONDS)
+                            curMeasureline.Time - beatlinesThisCountdown.Last().Time >= WaitCountdown.MIN_MEASURE_LENGTH)
                         {
                             beatlinesThisCountdown.Add(curMeasureline);
                         }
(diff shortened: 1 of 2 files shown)


YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -165,12 +165,12 @@ protected override void GenerateQueuedUpdates(double nextTime)
                 
                 if (State.IsWaitCountdownActive)
                 {
-                    double changeTime = currentCountdown.GetNextUpdateTime();
+                    double countdownEndTime = currentCountdown.TimeEnd;
 
-                    if (IsTimeBetween(changeTime, previousTime, nextTime))
+                    if (IsTimeBetween(countdownEndTime, previousTime, nextTime))
```
RileyTheFox, line 170: Currently the countdown ends at the exact time of the next note after the countdown's duration. Do we want to pull forward the end time a measure so that there's a measure of time where the countdown is no longer active and things like overstrums and overhits are now allowed?

My concern is that the game will be preventing overstrums/overhits at a time when the notes are hittable due to hitting windows in the front end early.

Purplo-cf:
Summary of changes:

- Overstrum/overhit detection is now turned back on when the fade-out animation starts.
- `State.IsWaitCountdownActive` now only represents the window where overstrum/overhit is ignored.
- The only timestamps being queued are a WaitCountdown's start time, and the fade-out time where `IsWaitCountdownActive` is set to `false`
- `State.CurrentWaitCountdownIndex` will only be incremented once the fade out animation has completed, so there will be a window of .45s where `IsWaitCountdownActive` will be set to `false` while the inactive countdown is still shown onscreen. `CurrentWaitCountdownIndex` is still being used to get progress bar percentage during this time.

I will get on the YARG-side changes this evening.

RileyTheFox (approved):
Waiting for https://github.com/YARC-Official/YARG/pull/825/ to be fixed up before merging as I will merge both at the same time

PR #197, opened 2024-07-20
Engine support for disabling authored SP phrases

Purplo-cf:
Necessary to disable SP phrases in Practice mode. This could also be used for a modifier down the line.

### DIFF
--- a/YARG.Core/Engine/BaseEngine.cs
+++ b/YARG.Core/Engine/BaseEngine.cs
@@ -441,11 +441,15 @@ namespace YARG.Core.Engine
             {
                 if (isAllowed)
                 {
-                    note.ResetNoteState();
+                    note.ResetFlags();
                 }
-                else
+                else if (note.IsStarPower)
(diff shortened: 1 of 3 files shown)


YARG.Core/Engine/BaseEngine.cs
```diff
@@ -426,6 +428,28 @@ protected void UpdateTimeVariables(double time)
...
+
+            foreach (var note in Notes)
+            {
+                if (isAllowed)
+                {
+                    note.ResetNoteState();
+                }
+                else
+                {
+                    StripStarPower(note);
```
RileyTheFox, line 448: Don't call this function as it requires modification to whether the phrase missed event should be called. Just directly strip the StarPower flag from every note (and child) here so StripStarPower is left unmodified.

RileyTheFox (changes requested):

RileyTheFox (approved):

PR #200, opened 2024-07-30
SP note gold effect core

Purplo-cf:
Color profile support for metal note edges in Rectangular theme. Starpower and non-Starpower variants

### DIFF
--- a/YARG.Core/Game/Presets/ColorProfile.cs
+++ b/YARG.Core/Game/Presets/ColorProfile.cs
@@ -17,6 +17,7 @@ namespace YARG.Core.Game
         public interface IFretColorProvider
         {
             public Color GetFretColor(int index);
+            public Color GetMetalColor(bool forStarPower);
             public Color GetFretInnerColor(int index);
             public Color GetParticleColor(int index);
         }
--- a/YARG.Core/Game/Presets/ColorProfile.Defaults.cs
+++ b/YARG.Core/Game/Presets/ColorProfile.Defaults.cs
@@ -17,7 +17,7 @@ namespace YARG.Core.Game
         private static readonly Color DefaultStarpower = Color.White; // #FFFFFF
 
         private static readonly Color DefaultNoteMetal = Color.White; // #FFFFFF
-        private static readonly Color DefaultNoteMetalStarpower = Color.FromArgb(0xFF, 0xE3, 0xBD, 0x00); // #E3BD00
+        private static readonly Color DefaultNoteMetalStarpower = Color.FromArgb(0xFF, 0xE3, 0xA8, 0x00); // #E3A800
 
         #endregion
 
--- a/YARG.Core/Game/Presets/ColorProfile.Defaults.cs
+++ b/YARG.Core/Game/Presets/ColorProfile.Defaults.cs
@@ -16,6 +16,9 @@ namespace YARG.Core.Game
 
         private static readonly Color DefaultStarpower = Color.White; // #FFFFFF
 
+        private static readonly Color DefaultNoteMetal = Color.White; // #FFFFFF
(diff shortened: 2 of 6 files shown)


Purplo-cf:
I was hesitant to add too many color profile options for something that is only visible in the Rectangular theme. In this first pass, I have only added two new color settings - one metal color for all normal notes, and one metal color for all SP notes.

If you prefer, I can make this more customizable so that a player can have different metal colors for every note, in and out of SP phrases.

TheNathannator:
Yeah I'm not really sure about adding color profile support for this yet, as currently it's specific to the rectangular note theme (at least in naming). Could be renamed to something like `Base` instead of `Metal` as a general base color for the note, so that this could be re-used for the circular theme as well.

In the end though, I don't know if we'll be able to maintain 100% compatibility for color profiles when custom notes become a thing. Different themes will want to do things in different ways (rectangular re-colors notes in SP phrases, while circular re-colors them when activated), so in an ideal world we can't quite have a "one size fits all" solution for storing customization. But this also depends on how much customization we want to allow, e.g. if we script how things are colored, or if we just have a 3-group system for base + color + aux (side/top).

All things to be determined down the road.

Purplo-cf:
Okay, if we don't need to pull the colors from the user profile, then I can just put the current gold color in the main PR as a constant and close this one. Sound good?

TheNathannator:
I'd be fine with that for now, yeah, so long as the circular theme isn't affected in an unintended way.

TheNathannator:
@EliteAsian123 @kaduwaengertner Any additional input on the thoughts here?

kaduwaengertner:
Please ignore the worst drawing skills ever but I think this would be a good way to have the colors set for notes so it works in both our provided themes

image

Allows for good customization and in the future with themes users would be able to have a good amount of colors to use on custom notes

YARG.Core/Game/Presets/ColorProfile.cs
```diff
@@ -17,6 +17,7 @@ public partial class ColorProfile : BasePreset, IBinarySerializable
         public interface IFretColorProvider
         {
             public Color GetFretColor(int index);
+            public Color GetMetalColor(bool forStarPower);
```
EliteAsian123, line 20: Keep in mind this interface is supposed to be used for frets only, not the notes.

EliteAsian123 (changes requested):

EliteAsian123:
> Please ignore the worst drawing skills ever

Literally better than any sketch I've drawn in my life.

Anyways, I'm not entirely happy with the color profile system as it is right now because there are differences between the rectangular and circle theme that makes it really hard to have a singular "simple" color profile that *just* changes the note colors. To get around this, in the theme authoring system there is currently an option to ignore the starpower color for the color of the notes (since in the circular theme, starpower notes stay the same color; the model just changes). This makes it a little easier to have a "simple" color profile, however, it does limit the customization a little bit.

What Kadu suggested is perfect for the maximum amount of customization, however, this means that if someone *just* wants to change the shade of blue (as an example), they'd definitely have to make multiple color profiles for each theme which is a pain.

I think the best solution is to have "simple color profiles" that *only* change the colors of each of the lanes (frets + notes), and "advanced color profiles" that allow you to change every single little aspect of the theme (like what Kadu's sketch shows). That might be a little over-complicated, however, it makes it less of a pain to transfer color profiles between themes if all you're doing is changing the note colors.

Thoughts?

kaduwaengertner:
Yeah I agree is not the best, think I can make some UI changes to help simplify the process but having some type of toggle between simple/advanced customization would be nice to have

Also did a color guide that we can utlize for the frets in a way that would work for both themes:

image
Both notes and frets having 4 colors total allowing for full customization

Ill make some changes in the UI for color profiles and we can discuss more on how we should have it

For this PR I think setting for now as just the gold works and we can update the profile customization later

PR #203, opened 2024-08-13
Button-based Star Power activation for Vocals

thomeval:


YARG.Core/Engine/Vocals/Engines/YargVocalsEngine.cs
```diff
@@ -60,6 +60,10 @@ protected override void MutateStateWithInput(GameInput gameInput)
 
                 OnSing?.Invoke(true);
             }
+            else if (action is VocalsAction.StarPower && CanStarPowerActivate)
+            {
+                ActivateStarPower();
```
EliteAsian123, line 65: For consistency, this should probably be laid out like it is for the guitar engine:
https://github.com/YARC-Official/YARG.Core/blob/2dc378f7e011fef1dfff052521f6ac478ebc3a9c/YARG.Core/Engine/Guitar/Engines/YargFiveFretEngine.cs#L69-L72

EliteAsian123 (changes requested):
One quick thing and this should be good to go

YARG.Core/Engine/Vocals/Engines/YargVocalsEngine.cs
thomeval: Done

PR #204, opened 2024-08-18
Refactor .chart parsing to use `YARGChartFileReader`

TheNathannator:
This ensures parsing behavior consistency between song scanning and chart loading, and fixes a possible issue where closing braces `}` being present in a text event would trip up the parser and make it fail to load.

To support this, I made some changes to some of the .ini classes to un-hard-code their expectance on being used for song.ini files, due to `YARGChartFileReader` relying on that infrastructure in its implementation and results. `YARGChartFileReader.ExtractModifiers` now takes a lookup of allowed modifiers as a parameter, and `IniSection` now takes a constructor argument for known modifiers to use for debug validation.

### DIFF
--- a/YARG.Core/IO/Ini/IniSection.cs
+++ b/YARG.Core/IO/Ini/IniSection.cs
@@ -7,14 +7,71 @@ namespace YARG.Core.IO.Ini
 {
     public sealed class IniSection
     {
+#if DEBUG
+        private static readonly Dictionary<Type, ModifierType> _validations;
+
+        static IniSection()
+        {
+            _validations = new()
+            {
+                { typeof(SortString), ModifierType.SortString },
+                { typeof(string), ModifierType.String },
+                { typeof(ulong), ModifierType.UInt64 },
+                { typeof(long), ModifierType.Int64 },
+                { typeof(uint), ModifierType.UInt32 },
+                { typeof(int), ModifierType.Int32 },
+                { typeof(ushort), ModifierType.UInt16 },
+                { typeof(short), ModifierType.Int16 },
+                { typeof(bool), ModifierType.Bool },
+                { typeof(float), ModifierType.Float },
+                { typeof(double), ModifierType.Double },
+                { typeof(long[]), ModifierType.Int64Array },
+            };
+        }
+
+        private void ThrowIfNot<T>(string key)
+        {
+            if (!knownModifiers.TryGetValue(key, out var mod))
+            {
+                throw new ArgumentException($"Dev: {key} is not a valid modifier!");
+            }
+
+            var type = typeof(T);
+            var modifierType = _validations[type];
+            if (modifierType != mod.Type &&
+                (modifierType == ModifierType.SortString) != (mod.Type == ModifierType.SortString_Chart) &&
+                (modifierType == ModifierType.String) != (mod.Type == ModifierType.String_Chart))
+            {
+                throw new ArgumentException($"Dev: Modifier {key} is not of type {type}");
+            }
+        }
+#endif
+
         private readonly Dictionary<string, List<IniModifier>> modifiers;
+
+#if DEBUG
+        private readonly Dictionary<string, IniModifierCreator> knownModifiers;
+#endif
 
         public int Count => modifiers.Count;
 
-        public IniSection() { modifiers = new(); }
-        public IniSection(in Dictionary<string, List<IniModifier>> modifiers)
+
+        // `knownModifiers` is always provided as an argument so other code doesn't have to do `#if DEBUG` guards
+        public IniSection(in Dictionary<string, IniModifierCreator> knownModifiers)
+        {
+            modifiers = new();
+#if DEBUG
+            this.knownModifiers = knownModifiers;
+#endif
+        }
+
+        public IniSection(in Dictionary<string, List<IniModifier>> modifiers,
+            in Dictionary<string, IniModifierCreator> knownModifiers)
         {
             this.modifiers = modifiers;
+#if DEBUG
+            this.knownModifiers = knownModifiers;
+#endif
         }
 
         public void Append(in Dictionary<string, List<IniModifier>> modsToAdd)
@@ -36,7 +93,7 @@ namespace YARG.Core.IO.Ini
         public bool TryGet(in string key, out SortString str, in SortString defaultStr)
         {
 #if DEBUG
-            SongIniHandler.ThrowIfNot<SortString>(key);
+            ThrowIfNot<SortString>(key);
 #endif
             if (modifiers.TryGetValue(key, out var results))
             {
@@ -59,7 +116,7 @@ namespace YARG.Core.IO.Ini
         public bool TryGet(in string key, out SortString str, in string defaultStr)
         {
 #if DEBUG
-            SongIniHandler.ThrowIfNot<SortString>(key);
+            ThrowIfNot<SortString>(key);
 #endif
             if (modifiers.TryGetValue(key, out var results))
             {
@@ -82,7 +139,7 @@ namespace YARG.Core.IO.Ini
         public bool TryGet(in string key, out string str)
         {
 #if DEBUG
-            SongIniHandler.ThrowIfNot<string>(key);
+            ThrowIfNot<string>(key);
 #endif
             if (modifiers.TryGetValue(key, out var results))
             {
@@ -96,7 +153,7 @@ namespace YARG.Core.IO.Ini
         public bool TryGet(in string key, out long val1, out long val2)
         {
 #if DEBUG
-            SongIniHandler.ThrowIfNot<long[]>(key);
+            ThrowIfNot<long[]>(key);
 #endif
             if (modifiers.TryGetValue(key, out var results))
             {
@@ -117,7 +174,7 @@ namespace YARG.Core.IO.Ini
             where T : unmanaged
         {
 #if DEBUG
-            SongIniHandler.ThrowIfNot<T>(key);
+            ThrowIfNot<T>(key);
 #endif
             if (modifiers.TryGetValue(key, out var results))
             {
--- a/YARG.Core/IO/Ini/IniSection.cs
+++ b/YARG.Core/IO/Ini/IniSection.cs
@@ -48,30 +48,22 @@ namespace YARG.Core.IO.Ini
 #endif
 
         private readonly Dictionary<string, List<IniModifier>> modifiers;
-
-#if DEBUG
         private readonly Dictionary<string, IniModifierCreator> knownModifiers;
-#endif
 
         public int Count => modifiers.Count;
 
 
-        // `knownModifiers` is always provided as an argument so other code doesn't have to do `#if DEBUG` guards
         public IniSection(in Dictionary<string, IniModifierCreator> knownModifiers)
         {
             modifiers = new();
-#if DEBUG
             this.knownModifiers = knownModifiers;
-#endif
         }
 
         public IniSection(in Dictionary<string, List<IniModifier>> modifiers,
             in Dictionary<string, IniModifierCreator> knownModifiers)
         {
             this.modifiers = modifiers;
-#if DEBUG
             this.knownModifiers = knownModifiers;
-#endif
         }
 
         public void Append(in Dictionary<string, List<IniModifier>> modsToAdd)
--- a/YARG.Core/IO/YARGChartFileReader.cs
+++ b/YARG.Core/IO/YARGChartFileReader.cs
@@ -240,10 +240,11 @@ namespace YARG.Core.IO
             return true;
         }
 
-        public static Dictionary<string, List<IniModifier>> ExtractChartModifiers<TChar>(ref YARGTextContainer<TChar> container)
+        public static IniSection ExtractChartModifiers<TChar>(ref YARGTextContainer<TChar> container)
             where TChar : unmanaged, IEquatable<TChar>, IConvertible
         {
-            return ExtractModifiers(ref container, METADATA_MODIFIERS);
+            var modifiers = ExtractModifiers(ref container, METADATA_MODIFIERS);
+            return new IniSection(modifiers, METADATA_MODIFIERS);
         }
 
         public static Dictionary<string, List<IniModifier>> ExtractModifiers<TChar>(
--- a/YARG.Core/IO/YARGChartFileReader.cs
+++ b/YARG.Core/IO/YARGChartFileReader.cs
@@ -77,29 +77,6 @@ namespace YARG.Core.IO
             new("S",  ChartEventType.Special),
         };
 
-        public static readonly Dictionary<string, IniModifierCreator> METADATA_MODIFIERS = new()
-        {
-            { "Name",           new("Name", ModifierType.String_Chart) },
-            { "Artist",         new("Artist", ModifierType.String_Chart) },
-            { "Album",          new("Album", ModifierType.String_Chart) },
-            { "Genre",          new("Genre", ModifierType.String_Chart) },
-            { "Year",           new("Year", ModifierType.String_Chart) },
-            { "Charter",        new("Charter", ModifierType.String_Chart) },
-
-            { "Difficulty",     new("Difficulty", ModifierType.Int32) },
-            { "Length",         new("Length", ModifierType.Double) },
-            { "Offset",         new("Offset", ModifierType.Double) },
-            { "PreviewStart",   new("PreviewStart", ModifierType.Double) },
-            { "PreviewEnd",     new("PreviewEnd", ModifierType.Double) },
-
-            { "Resolution",     new("Resolution", ModifierType.UInt32) },
-        };
-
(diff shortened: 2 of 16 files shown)


YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
```diff
@@ -277,129 +174,137 @@ private static void ValidateAndApplySettings(MoonSong song, ref ParseSettings se
...
+                        case ChartEventType.Time_Sig:
+                        {
+                            uint numerator = YARGTextReader.ExtractUInt32AndWhitespace(ref chartText);
 
-                    tempoTracker.Update(tick);
+                            // Denominator is optional, and defaults to 2 (becomes 4)
+                            if (!YARGTextReader.TryExtractUInt32(ref chartText, out uint denominator))
+                                denominator = 2;
+                            else
+                                YARGTextReader.SkipWhitespace(ref chartText);
```
sonicfind, line 217: Skipping the whitespace shouldn't be necessary since it's going to skip to the next line in the following `TryParseEvent` call.

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
```diff
@@ -277,129 +174,137 @@ private static void ValidateAndApplySettings(MoonSong song, ref ParseSettings se
...
+                        case ChartEventType.Anchor:
+                        {
+                            // Ignored for now, we don't need anchors
+                            // ulong anchorTime = YARGTextReader.ExtractUInt64AndWhitespace(ref chartText);
+                            // var anchor = new Anchor()
+                            // {
+                            //     tick = tick,
+                            //     anchorTime = anchorTime / 1000000.0
+                            // };
+                            // anchorData.Add(anchor);
```
sonicfind, line 236: We may have to reconsider *this* later, but for now it's fine. It'd go outta scope for the PR.

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
```diff
@@ -277,129 +174,137 @@ private static void ValidateAndApplySettings(MoonSong song, ref ParseSettings se
...
-                    // Split on the equals sign
-                    var tickText = line.SplitOnceTrimmed('=', out var remaining);
-
-                    // Get tick
-                    uint tick = (uint) FastInt32Parse(tickText);
-
-                    if (prevTick > tick) throw new Exception("Tick value not in ascending order");
+                    uint tick = (uint) chartEvent.Position;
+                    if (prevTick > tick)
+                        throw new Exception("Tick value not in ascending order");
```
sonicfind, line 266: Again, `TryParseEvent` throws in this case

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
```diff
@@ -418,108 +323,85 @@ private static void LoadChart(ref ParseSettings settings, MoonSong song, AsciiTr
...
-
-                        // Get tick
-                        uint tick = (uint) FastInt32Parse(tickText);
-
-                        if (prevTick > tick) throw new Exception("Tick value not in ascending order");
-                        prevTick = tick;
+                    uint tick = (uint) chartEvent.Position;
+                    if (prevTick > tick)
+                        throw new Exception("Tick value not in ascending order");
+                    prevTick = tick;
```
sonicfind, line 335: Another one for the road

YARG.Core/IO/YARGChartFileReader.cs
```diff
@@ -217,14 +240,22 @@ public static unsafe bool TryParseEvent<TChar>(ref YARGTextContainer<TChar> cont
             return true;
         }
 
-        public unsafe static Dictionary<string, List<IniModifier>> ExtractModifiers<TChar>(ref YARGTextContainer<TChar> container)
+        public static IniSection ExtractChartModifiers<TChar>(ref YARGTextContainer<TChar> container)
+            where TChar : unmanaged, IEquatable<TChar>, IConvertible
+        {
+            var modifiers = ExtractModifiers(ref container, METADATA_MODIFIERS);
+            return new IniSection(modifiers, METADATA_MODIFIERS);
+        }
```
sonicfind, line 248: We only call this from one location, so adding a special function for that case just to wrap the IniSection construction is unnecessary. Just let the caller handle that detail and only have the two-parameter version. Makes it more explicit.

sonicfind (changes requested):

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
TheNathannator: Fair point, I was just ensuring consistent behavior with the other value extraction calls.

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
```diff
@@ -90,184 +90,81 @@ public static MoonSong ReadFromText(ReadOnlySpan<char> chartText)
...
-                return false;
-
-            if (nameEndIndex < 0)
-            {
-                int startLine = GetLineCount(chartText, index, nameStartIndex);
-                throw new Exception($"Missing end bracket for section name on line {startLine}!");
-            }
-
-            if (nameEndIndex < nameStartIndex)
+            static void ExpectSection(ref YARGTextContainer<TChar> chartText, string section, int sectionIndex)
```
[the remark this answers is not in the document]
TheNathannator: In practice I'm not sure if this case will ever actually come up as an issue; the resolution and sync track are vital to the chart working correctly, and I'd rather reject charts that are missing these than cover for their mistakes.

The only case this will happen in the wild is is if someone deliberately removed those sections. Moonscraper will always write `[Song]`, `[SyncTrack]`, and `[Events]`, in that order, and I presume Feedback did as well.

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
TheNathannator: We'll get to it when we're ready for it, no point parsing in something that won't be used until an indeterminate timeframe lol

YARG.Core/IO/YARGChartFileReader.cs
TheNathannator: I was thinking from an API design perspective with this one, having the default is convenient. But yeah, it's a little superfluous when it's used in just one place.

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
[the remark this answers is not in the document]
sonicfind: The issue is it's hard to call the deliberate removal of those sections and/or the resolution "mistakes" when there really aren't *hard* guidelines or specifications regarding how to handle those situations. In absence of that kind of stuff, they only thing we can go off of is precedence; and currently, the precedence set by the most prevalent and used tools/games that accept the format is to just use default values instead of erroring. 120 bpm with 4/4 time sig for a missing synctrack (which is already handled in MoonSong's construction in the refactor branch) and 192 for a missing resolution value.

Plus you said it yourself: it'd need to be *deliberate*, and they'd only do that if the file still worked.

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
```diff
@@ -90,184 +90,81 @@ public static MoonSong ReadFromText(ReadOnlySpan<char> chartText)
...
-            throw new InvalidDataException("No resolution was found in the chart data!");
+            // Resolution = 192
+            if (!metadata.TryGet("Resolution", out uint resolution))
+                throw new InvalidDataException("Could not read .chart resolution!");
+            if (resolution < 1)
+                throw new InvalidDataException($"Invalid .chart resolution {resolution}! Must be non-zero and non-negative");
+
+            var song = new MoonSong(resolution);
+            ValidateAndApplySettings(song, ref settings);
+            return song;
```
sonicfind: You can simplify this function a bit. The main point of importance is that it parses (or attempts to parse) the tick rate of the chart. All the other stuff should be done externally - especially the `ValidateAndApplySettings` call. Just have the function return only the tick rate.

Additionally, going off the whole "not an error" thing, you can move the `[Song]` header check into this function instead. This would allow easily handling the default case. You'd define the tick rate to the default .chart value of 192, then only if the header is found and if `Resolution` rests within it would you override it. Though, I'd still recommend throwing for a resolution of `< 1`, because that'd be stupid.

YARG.Core/IO/Ini/IniSection.cs
```diff
@@ -7,14 +7,71 @@ namespace YARG.Core.IO.Ini
...
+            }
+        }
+#endif
+
         private readonly Dictionary<string, List<IniModifier>> modifiers;
 
+#if DEBUG
+        private readonly Dictionary<string, IniModifierCreator> knownModifiers;
+#endif
+
```
sonicfind, line 55: Don't wrap this (or its initialization) as "Debug only". I might have a use for this afterwards.

sonicfind (changes requested):

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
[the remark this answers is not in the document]
TheNathannator: I'd much rather have the resolution and sync track be guaranteed to be correct ahead of time, than have other things which might rely on it during parsing have incorrect results due to absence or ordering differences. This might not apply now, but it certainly could in the future with potential refactors.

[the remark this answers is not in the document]
TheNathannator: To equate this to a prior situation: my reasoning for this is the same reason as to why we decided to disallow having multiple track names within a MIDI track. It's an ambiguous scenario with no real clear answer other than "set some defaults and hope for the best". Just because it *can* happen doesn't mean we need to accept it, regardless of prior precedence.

To contrast, that change affected quite a few charts out there in the wild, while I expect this to affect no charts whatsoever due to the obscurity of the problem. Allowing it would only cause more potential problems, imo.

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
TheNathannator: > especially the `ValidateAndApplySettings` call.

This call relies directly on values read from the metadata section, so IMO it doesn't make much sense to separate the two. This function is as simple as it needs to be, no need to make it simpler just for the heck of it.

sonicfind: In actuality, it relies on only two things: resolution/tick rate & the `settings` variable. The tick rate can just be returned and `settings` doesn't need to be passed in.

The problem right now is that the name `ReadMetadataSection` doesn't actually match up with what is transpiring in the function.
1. It does read the metadata segment. That's fine. However...
2. It only extracts out the tick rate from the entirety of that section. That detail isn't apparent to a caller through the name.
3. It creates a `MoonSong` object, which leads to the initialization of a bunch of data and separate states. Again, that detail is not apparent in the function name.
4. It then uses the extracted tick rate to modify state in the song and, more importantly, the settings object. Again, not apparent.

Unfortunately, so long as all that stuff remains stays in this one function, there's no sensible way to solve any of those clarity issues without making the function name like one you would see in a weird-ass Java codebase (those bitches are longggggggg).

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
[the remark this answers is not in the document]
sonicfind: > To equate this to a prior situation: my reasoning for this is the same reason as to why we decided to disallow having multiple track names within a MIDI track. It's an ambiguous scenario with no real clear answer other than "set some defaults and hope for the best"

The big difference here is that MIDI has a hard specification, and said specification explicitly states that a midi event of type `03` represents "**the** name of the track", not "**A** name for the track". Therefore, tracks contained in midi files are only in-spec if they contain no more than a single track name. `.chart` doesn't really have an equivalent to midi's specification. It has documentation on the format, but that's from research of how it's been used, not how it's supposed to be used.

And unlike this where these situations are mainly caused through deliberate actions or decisions, the duplicate MIDI track name issues are made through **accident**. They're not the same.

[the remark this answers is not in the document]
TheNathannator: They're still a very similar caliber of issue, disregarding standards/documentation. It's a behavior that's colloquially allowed by prior programs/games which we would now be deciding to disallow for correctness.

I was initially against the decision to disallow these multiple track names in .mid, because it was very much a thing that already existed in notable charts. But my opinion was outweighed, and so I conceded, despite it being a compatibility-breaking change.

This change does not have the same compatibility implications. There is no reason for us to allow this, other than the fact that previous games allow it. No existing charts I'm aware of rely on this behavior to function correctly; other than that one chart with a quoted resolution that I pointed out in the Discord (which we can account for in other, more correct ways).

Even more to the point, this being a very deliberate action is even more reason *not* to support it, while the .mid issue being accidental and on a somewhat notable scale would be grounds enough for me to want to implement a workaround, rather than require charts to be fixed up manually.

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
TheNathannator: > That detail isn't apparent to a caller through the name.

This is also a private method, whose details probably *should* be looked at if someone is analyzing the code in any level of detail, IMO.

I consider the reading of the resolution value to still be a detail related to the metadata section. The fact that *only* resolution is read doesn't particularly matter, because most of the remaining properties aren't relevant to the actual parsing of the chart (and the ones that do matter are already handled at scan time, through `ParseSettings`).

- By that logic it's a little superfluous to include *all* the properties in `YARGChartFileReader.METADATA_MODIFIERS`, and so I'm fine with removing the unnecessary ones.
- Side-note: I find it odd how .chart properties are also included in the map of song.ini properties in the scanning code. It makes it less-than-obvious on how those are converted to the corresponding song.ini properties; but that's something for a separate discussion.

The creation of a `MoonSong` object relies directly on the resolution value, and so the easiest place to stick its creation was in there. I do agree that part is odd though, so I'll move it. Returning *just* the resolution doesn't make the most sense with a method that's supposed to contain details though, so I'm thinking a struct would be better/clearer (even if it's just got one member currently).

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.cs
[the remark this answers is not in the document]
sonicfind: > Even more to the point, this being a very deliberate action is even more reason not to support it

What's important here is *why* someone would deliberately remove the `[Song]` section. Think about how Moonscraper exports .chart files. Every time, the `[Song]` section is filled with different metadata: song name, artist, album, year, offset, Resolution, Player2, Difficulty, PreviewStart, PreviewEnd, genre, MediaType, and any music streams.

How many of those modifiers have any actual use in any games? You could argue that the basic song info metadata is... but that's what we use song.ini for. And neither CH or YARG utilize the stream paths (because that'd be a huge can of worms). It wouldn't surprise me if someone decided to save just a little bit more room by removing any of that duplicate and/or unused information.

That leaves the resolution. Well, for CH, even before reading this section, the resolution is already defaulted to 192. With that being the case, of course there would be no harm in just stripping that out too. So now all that's left is... nothing? Well, might as well removing that entire section then! It still works in CH! Same with synctrack... although that's only useful for a song that is entirely 120bpm and 4/4 (exceedingly rare - unless you like Big And Chunky).

It's purely a game of bytes, saving as much space as you can manage while still being a valid playable chart in the main most popular five-fret fan game in community. These *have a purpose*, and so long as a well-defined explicit specification does not exist for the format, longstanding precedence rules.

> while the .mid issue being accidental and on a somewhat notable scale would be grounds enough for me to want to implement a workaround

Unlike .mid, where there does exist a well-defined explicit specification created and deliberated by numerous groups of important people and corporations; a specification designed to provide guidelines on how midi files have to be formatted to ensure fully consistent behavior across the many tools that accept them.

For the majority of times we've seen people ask about the track name restriction, it's the result of an accidental misunderstanding of Reaper to some degree. As a result, they gain new information about the format as they aim to fix it, and in the end, the file usually ends up tidier (bonus: **less data!**). Hiding the error would only lead to people never learning correct patterns, which can have some very adverse effects. I'd rather not risk having multiple instances of this happening under someone's nose.
image

The only midi spec violation we must allow is ignoring sysex for MidiNote resets: because, unfortunately, besides Phase Shift, some official Rock Band midi files contain that error (and I mean on disc and/or direct DLC). It'd be *really* bad to not allow those to work. That being said, no one uses tools that violate that on export these days, so it's not as big of a deal.

These are two wildly different situations with massively different implications.

PR #206, opened 2024-08-24
Fix solo & solo_end .chart conversions

sonicfind:


YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.ProcessLists.cs
```diff
@@ -201,31 +201,40 @@ private static List<NoteEventProcessFn> GetInitialPostProcessList(MoonChart.Game
...
-
-            public SoloPhraseConverter(MoonChart chart)
-            {
-                _chart = chart;
-            }
-
-            public void AddPhrase(uint startTick, uint endTick)
+            var chart = noteProcessParams.chart;
+            // Keeps tracks of soloes that start on the same tick when another solo ends
+            var soloQueue = stackalloc uint[2] { uint.MaxValue, uint.MaxValue };
```
TheNathannator, line 208: No real need to stackalloc here if you're not using this as an actual array or buffer, it could just as easily be two separate variables. I feel the intent becomes obscured otherwise:
```suggestion
            uint startTick = uint.MaxValue;
            uint nextStartTick = uint.MaxValue;
```

TheNathannator (approved):
Might be good to have a test for this, otherwise this should be good.

PR #208, opened 2024-09-01
Alternate Take On A Replay Refactor

sonicfind:
Mainly to provide a contrasting perspective (unless we decide to roll with it):

Completely overhauls the replay file format and how the app serializes it.

Firstly, one of the described purposes on the replay versioning system was that old replays, even if out of date, to some degree, should stay legible. That is, to say, that minimally, we should be able to decipher just enough information that we can add the replay as an entry to the in-game history or gamerecords menus. This was a bit problematic beforehand as the types that handled these contained shared conflicting information. You had to first attempt to read the *entire* file - which might not even work because the old lack of proper version control -, then extract just that metadata info out to create the entry type.

With this update, I've completely separated the "metadata" information into a different structure (`ReplayInfo`) from the actual data controlling the replay (`ReplayData`). This means that the metadata can exist independently from the replay data that it maps to, which grants a bit more freedom in regards how we decide to de-serialize replay files in certain situations.

Replay de-serialization thereby gets a new "mode": metadata only. When cataloging a collection of replays, the actual replay data should only be used for simple validation against the hash. Therefore, a separate function is provided that does just that.

Secondly, the replay file has seen a major update:
1. The starting header tag has changed from YARGPLAY to YAREPLAY. This allows us to easily distinguish old replays from new ones right out the gate, allowing pre-version 6 replays to immediately communicate their "metadata only" state.
2. The replay metadata and main data now have two separate hashes. The hash for the metadata is provided at the start of the file (alongside the length of the metadata chunk) while the hash for the main replay data rests *within* the metadata chunk.
3. Additionally, the engine & replay version values have been moved into the metadata chunk as well. This adds another step of security as the hashing of the metadata will therefore disallow any changes to those values after the fact.
4. Added a header to tag to every `ReplayFrame` chunk. It provides an extra simple mechanism for validating data integrity.

Thirdly, the serialization and de-serialization of the actual replay data has taken a more self-sufficient approach. Each structure, when appropriate, will take turns constructing themselves from the stream of the file - and vice verse with serializing *to* a file. The reason for this is simple: most of the data that we write and read to and from a file is supposed to maintain a `readonly` state. However, the state of the prior handling of the data did not make that intention clear as explicitly marking the values as `readonly` was not possible. Moving to this more-object based approach now allows that, making the logical flow of the entire process easier to grasp.

+ Additional note:
  + Removed engine logger types entirely (as per Riley's recommendation)
  + Added functions to StreamExtensions (some from BinaryReaderWriterExtensions)
  + REQUIRES MAIN REPO UPDATE


YARG.Core/Replays/Analyzer/ReplayAnalyzer.cs
```diff
@@ -113,7 +108,6 @@ private AnalysisResult RunFrame(ReplayFrame frame)
             {
                 Passed = passed,
                 Stats = engine.BaseStats,
-                EventLogger = _keepEngineLoggers ? engine.EventLogger : null,
             };
         }
```
RileyTheFox, line 113: Need to configure the ReplayAnalyzer for Pro Keys like I did in [this commit](https://github.com/YARC-Official/YARG.Core/pull/180/commits/a1baf369ca1b7b4fd91cfe12977680155dcc850f)

RileyTheFox (changes requested):
Other than the one comment, looks good to me

YARG.Core/Replays/Analyzer/ReplayAnalyzer.cs
sonicfind: aight

RileyTheFox (approved):

PR #211, opened 2024-09-10
Fix Drum SP Activation notes being mandatory

thomeval:
This PR fixes an inssue that causes drum SP activation notes to be mandatory. These notes should be skippable, allowing drummers to save SP for later without penalty.


EliteAsian123 (changes requested):
Hey, sorry for the late reply!

On the master branch, this logic is in the `YargDrumsEngine` as opposed to in the `MissNote` method: https://github.com/YARC-Official/YARG.Core/blob/985e3d27e2936295c64b43135bff7c458f8ab5d7/YARG.Core/Engine/Drums/Engines/YargDrumsEngine.cs#L92

If you can move the logic there that'd ensure consistency and also be a little bit less confusing. Thanks!

thomeval:
Okay, ready for another review.

PR #218, opened 2024-10-06
Countdown rework core

Purplo-cf:
Changed engine countdown events to be purely time-based


RileyTheFox:
Going to keep this one on hold for a little while as I work on a lot of other engine fixes, and I'll just merge them all as one.

Reason being is that this changes how countdowns are generated (start and end times), which will affect the overstrum prevention which requires an engine version change for replays. So I'm just going to do it all in one go

Purplo-cf:
Good call. Trill and Tremolo lanes are close to ready and those will also be a replay breaking change

PR #220, opened 2024-10-18
Customizable Highway Presets

thomeval:
This PR introduces Highway Presets to YARG.Core. For details, see the associated PR on the YARG repo.


EliteAsian123 (changes requested):
Now, please let me know if I'm wrong here, however, I think it's actually better to handle highway presets much like themes are in this case: don't store them in replays (only the GUID) and move the class to YARG instead of YARG.Core. 

The reason why is because these will eventually be expanded to include textures (or file paths to textures) so you can customize the highways even more. In that case, storing the highway preset in the replay doesn't make sense since everyone wouldn't have said texture at said path. This is essentially the same reason why themes aren't store in replays (only the GUIDs so if the user _has_ the theme, then it's applied).

Obviously, it is possible that in the future we will store them in replays, however, then there's an explicit content risk, and potentially other risks. Because of this I think it's better to cross that bridge when we get to it, and after we discuss it a little more.

thomeval:
I've removed HighwayPresets from ReplayData as requested.

YARG.Core/Game/Presets/HighwayPreset.cs
```diff
@@ -0,0 +1,81 @@
...
+            writer.Write(StarPowerColor);
+            writer.Write(BackgroundBaseColor1);
+            writer.Write(BackgroundBaseColor2);
+            writer.Write(BackgroundBaseColor3);
+            writer.Write(BackgroundPatternColor);
+            writer.Write(BackgroundGrooveBaseColor1);
+            writer.Write(BackgroundGrooveBaseColor2);
+            writer.Write(BackgroundGrooveBaseColor3);
+            writer.Write(BackgroundGroovePatternColor);
+        }
```
EliteAsian123, line 79: This shouldn't be needed I believe, since they're not used in replays

EliteAsian123 (changes requested):
Hello! Sorry for the late review. Just some changes and this should be good to go! I'll review the other PR shortly.

PR #222, opened 2024-10-27
Lanes core

Purplo-cf:
Parsing and engine support for trill and tremolo lanes on Guitar, Drums, and trill lanes on Pro Keys.

### DIFF
--- a/YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Drums.cs
+++ b/YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Drums.cs
@@ -9,6 +9,9 @@ namespace YARG.Core.Chart
     {
         private bool _discoFlip = false;
 
+        private uint _lastTremoloTick;
+        private List<int>? _validTremoloNotes = null; 
+
         public InstrumentTrack<DrumNote> LoadDrumsTrack(Instrument instrument)
         {
             _discoFlip = false;
@@ -37,8 +40,13 @@ namespace YARG.Core.Chart
         {
             var pad = GetFourLaneDrumPad(moonNote);
             var noteType = GetDrumNoteType(moonNote);
+
             var generalFlags = GetGeneralFlags(moonNote, currentPhrases);
-            generalFlags |= GetTremoloFlags(moonNote, currentPhrases, isDrumNote:true);
+            if (currentPhrases.TryGetValue(MoonPhrase.Type.TremoloLane, out var tremolo))
+            {
+                generalFlags = ModifyTremoloFlags(moonNote, generalFlags, tremolo);
+            }
+
             var drumFlags = GetDrumNoteFlags(moonNote, currentPhrases);
 
             double time = _moonSong.TickToTime(moonNote.tick);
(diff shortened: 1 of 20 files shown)


YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Drums.cs
Purplo-cf, line 1: For Guitar and Keys, every note inside of a trill and tremolo phrase is assumed to be part of the lane. There are special considerations for drum rolls and cymbal swells where that is not always the case. Only the notes that appear the most in the phrase will get the appropriate lane flag.

Chord tremolos are still possible here if more than one note is tied for the most appearances, but kick notes will not be included in the case of a tie. Kick tremolos are only possible when kick notes are the undisputed winner.

YARG.Core/Chart/Notes/Note.cs
Purplo-cf, line 1: `Note.LaneNote` is read during generic `BaseEngine` and `TrackPlayer` logic to get around instrument-specific property names. `GuitarNote` returns `NoteMask` to allow for tremolo chords.

YARG.Core/Engine/BaseEngine.cs
Purplo-cf, line 1: `TotalLanes`/`CurrentLaneIndex` keep track of how many lanes are left at the current time in the chart. If there aren't any left, then lane logic is skipped.

`RequiredLaneNote` is the next expected player input to extend the leniency window while lane behavior is active. This stays constant during tremolo phrases, and alternates between two values during trill phrases.

`NextTrillNote` and `RequiredLaneNote` swap values every time a player input satisfies the current `RequiredLaneNote` value.

sonicfind: I don't handle the engine stuff, but just a suggestion: put this information *in the struct* as summary comments.

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -374,6 +396,65 @@ protected virtual void HitNote(TNoteType note)
...
+            {
+                // Lane ends with this note, continue overstrum leniency while transitioning out of this lane
+                YargLogger.LogFormatTrace("Lane ending at {0}", CurrentTime);
+
+                UpdateLaneExpireTime();
+            }
+        }
+
+        // Intercept a missed note while a lane phrase is active with allowances remaining
+        protected bool HitNoteFromLane(TNoteType note)
```
Purplo-cf, line 435: A check for `HitNoteFromLane` should be included anywhere that `MissNote` can be called across all engines. This will call `HitNote` on a note that is about to be missed or skipped if it that note is part of the active lane phrase

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -384,12 +465,83 @@ protected virtual void MissNote(TNoteType note)
             }
         }
 
+        protected void SubmitLaneNote(int newNote)
```
Purplo-cf, line 468: `SubmitLaneNote` is called from `MutateStateWithInput` across all engines.

YARG.Core/Engine/HitWindowSettings.cs
Purplo-cf, line 1: `TremoloFrontEndPercent` determines how much to increase `LaneExpireTime` when the active lane is satisfied. This is expressed as a percentage of the current maximum frontend size. A value of 1.0 will forgive every note in the hit window. Larger values will extend the forgiveness threshold even further.

PR #224, opened 2024-11-27
Engine support for treating solo frets separately from normal frets

wyrdough:
This PR adds support for solo buttons as found on Rock Band guitars and the Riffmaster.

During a solo section, all notes are treated as if they are tap notes when the solo buttons are used.


RileyTheFox:
We decided a while ago that the change in logic solo buttons provide was something we were not going to have as it provides an advantage over guitars which don't have solo buttons.

We've considered this being added as some kind of modifier/engine setting but it's not been thought about too much.

wyrdough:
Changed to an engine parameter. Logic changes as suggested on Discord to prevent tap logic being triggered while standard buttons are held.

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -358,11 +363,31 @@ protected sealed override int CalculateBaseScore()
 
         protected void ToggleFret(int fret, bool active)
         {
+            // FIXME: This is a terrible hack
+            if((int) fret >= 10) {
```
EliteAsian123, line 367: Please please please do not use "terrible hacks" within PRs. 

Before the big rewrite of v0.12, I made the mistake of merging PRs with hacks in them. They build up over time, and the codebase's quality just plummets because of that. 

Unfortunately, because of this, before this PR is merged this needs to be fixed.

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -358,11 +363,31 @@ protected sealed override int CalculateBaseScore()
...
+                    // Subtract one from count, avoiding underflow
+                    StandardButtonCount = (StandardButtonCount > 0) ? StandardButtonCount - 1 : 0;
+                }
+            }
             ButtonMask = (byte) (active ? ButtonMask | (1 << fret) : ButtonMask & ~(1 << fret));
         }
 
         public bool IsFretHeld(GuitarAction fret)
         {
+            // FIXME: This is a terrible hack
```
EliteAsian123, line 386: Same here

EliteAsian123 (changes requested):
Hello, thanks for the PR!

Please review the comments below for the changes I requested.

After that fix, I just need to know if this conflicts with any changes in the `engine-fixes` branch and whether or not this can be merged right now, or if we should wait (@RileyTheFox).

YARG.Core/Engine/Guitar/GuitarEngine.cs
wyrdough: That is there so that the solo frets don't have to be special cased everywhere ButtonMask is used. I called it a hack, but I can't really think of a better way to do it, TBH.

wyrdough: Ok, I changed it to make the input button mask a ushort to accommodate the extra buttons and then calculate an effective button mask which is a bitwise or of the two sets of buttons to be used where ButtonMask was previously being used so that the solo buttons still act as regular buttons outside of solos.

If you can think of a better strategy here, I'm all ears.

PR #227, opened 2025-01-02
Remove logging of countdowns not added

wyrdough:
This fixes a bug causing some songs to fail to initialize due to an ArgumentOutOfRange exception in the call to YargLogger.


RileyTheFox:
Honestly just remove the logs for not creating a countdown altogether. Don't really see the point in them.

wyrdough:
Ok. Done.

PR #228, opened 2025-01-03
Clamp measure index at 0 to avoid exception when no events at time 0

Nyxyxylyth:
If the first event on the BEAT track does not occur at time 0, a failure to load would occur with a System.ArgumentOutOfRangeException, because the -1 returned by GetIndexOfPrevious would then be used as an array index.  Clamp the value at 0 to use the time of the first event instead.

Fixes the issue detailed in Discord bug-reports: https://discord.com/channels/1086048856678084609/1324542191556759662


Nyxyxylyth:
Another bug report, same issue: https://discord.com/channels/1086048856678084609/1325249930893328515

PR #229, opened 2025-01-11
Handle unison phrases

wyrdough:
This adds an EngineManager class that will eventually award unison star power bonuses, track band multipliers, and whatever else can't be done within a single engine.

Remaining tasks:

- [x] Actually award bonus star power on successful unison completion.
~~- [ ] Track band score separately from individual scores~~
~~- [ ] Track band multipliers and award extra band score~~
- [x] Rework replay code as necessary to accommodate star power bonuses

Since this doesn't involve changes to the engines themselves, I don't believe this needs to be based on engine-fixes.


wyrdough:
Removed the band scoring goal for now, as it requires more consultation and time in the oven.

PR #230, opened 2025-01-25
Move practice section parsing to YARG

Frickitickitavi:
**Requires YARG PR 950**

Removes transformation logic from practice section parsing. My other PR moves that logic to YARG, where it can take advantage of localization. YARG.Core now simply checks whether the event begins with `prc` or `section` and, if so, passes it unedited to YARG.

### DIFF
--- a/YARG.Core/Parsing/TextEvents.cs
+++ b/YARG.Core/Parsing/TextEvents.cs
@@ -89,7 +89,6 @@ namespace YARG.Core.Parsing
         /// <returns>
         /// True if the event was parsed successfully, false otherwise.
         /// </returns>
-        // Equivalent to reading the capture of this regex: (?:section|prc)[ _](.*)
         public static bool TryParseSectionEvent(ReadOnlySpan<char> text, out ReadOnlySpan<char> name)
         {
             name = ReadOnlySpan<char>.Empty;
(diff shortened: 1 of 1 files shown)


TheNathannator:
I'm not sure I'm a fan of moving this part of the parsing outside of Core. What in particular do you need this moved out for?

Frickitickitavi:
> I'm not sure I'm a fan of moving this part of the parsing outside of Core. What in particular do you need this moved out for? Moving this out now means that anything which uses Core needs to implement this handling manually, which is not ideal.
> 
> IMO section names should be localized regardless of if they use `[section ...]` or `[prc_...]`, as both formats can contain names which need to be localized (e.g. everything before RB3 uses `[section ...]`).

I moved the logic to YARG because it makes use of `Localize` which, if I'm not mistaken, is not accessible in YARG.Core. Is there a way we can perform localization within Core? If not, where do you suggest drawing the line between Core and YARG's responsibilities? Does Core turn `[prc_gtr_solo_2a]` into `GTR SOLO 2A` and then pass that to YARG, which checks if that's a localized section name?

RileyTheFox:
> Is there a way we can perform localization within Core?
No, and you don't want to localise in core. You just want to load the practice sections into the song data as is. Those strings are then accessed by YARG when displaying a section name somewhere and then that checks if it should be localised. Core does not care about localisation as its all backend code. YARG is the front end which is responsible for localising.

TLDR: Core just needs to load the section string as is from the chart data. Check for the `section`/`prc` patrs in YARG and localise there

Frickitickitavi:
> > Is there a way we can perform localization within Core?
> 
> No, and you don't want to localise in core. You just want to load the practice sections into the song data as is. Those strings are then accessed by YARG when displaying a section name somewhere and then that checks if it should be localised. Core does not care about localisation as its all backend code. YARG is the front end which is responsible for localising.
> 
> TLDR: Core just needs to load the section string as is from the chart data. Check for the `section`/`prc` patrs in YARG and localise there

If I'm reading this correctly, it sounds like you agree with the current state of this PR? (not withstanding the issues identified in the YARG PR)

TheNathannator:
What I'm saying is this part of Core doesn't need to be changed to enable localization of section names. Both `[section gtr_solo]` and `[prc_gtr_solo]` should be localized identically.

Frickitickitavi:
Ohh, got it. I wasn't aware `section`-based events were ever used with the same contents as `prc` ones - I thought it was always plaintext like `[section Guitar Solo 1]` or whatever.

With that context, I see why we wouldn't want to change the logic here. Closing this PR and will update the YARG PR accordingly

PR #233, opened 2025-02-03
Scanning Refactor V3

sonicfind:
SongEntry:
+ Contain the `SortString` variants for name, artist, album, etc. in the type directly, but privately. These are set through a dedicated protected method after the lowest level deserialization step and after a successful new entry scan.
+ Hold a boolean value that communicates the `duplicate entry` status when sorted
+ `Location` -> `SortBasedLocation`; `DirectoryActual`  -> `ActualLocation`.
+ Simplistic constructors across the board. All (de)serialization is designated to specific functions called after the initial construction of an entry.
+ Utilize the newer flag-property of the `DrumsType` enum to implement a fallback hierarchy. If the `FourLane` bit is active, default to that. Otherwise, `ProDrums` if active, then finally `FiveLane`. This style simplifies handling the differences between the default behavior of .chart and .mid.

Cache:
+ No longer remove duplicate songs. Instead, keep the determined non-duplicate entry in any set at the front of its set and mark all others. This makes it so we don't require a full re-scan to account for changes to that setting. The music library will ask for that state in comparison to the setting to determine whether to add an entry to the visible list.

+ Remove separation of Parallel & Sequential style. We never end up using the sequential style, so it just ended up as code bloat.
+ Reorganize all cache groups. Packed CONs split the upgrades and entries into completely separate objects.
+ Apply all updates and upgrades as the last step *including* for entries read from cache during a full scan.
+ Only check for the .mid for a CON update entry. Since that's the only part that matters for determining the validity of an entry, there's no point writing the info into cache. Just have to track the base update directory and infer the proper location for the other items.
+ Utilize a specialized Enumerable struct and optimized cache format to more easily parallelize cache deserialization. Have every section follow a `Count: Length: (data)` format regardless of type. The only exceptions are the cache metadata strings where we predetermine the count.

Extraneous:
+ Add value-based `FixedArrayStream` that dodges the managed heap overhead of all the slices created during the de-serialization of cache with a `FixedArray<byte>`. (Note: also used for replays).
+ Attempt to constrain and encapsulate `unsafe` block usage as much as possible for certain types. ex: all the text reader stuff.
+ General changes to `CONFilestream` & `SNGFilestream` buffer loading


EliteAsian123:
Absolutely insane stuff. Great work!

PR #234, opened 2025-02-04
(Rebased) Fix and refactor chart parsing behavior tests; implement Pro Keys parsing test

TheNathannator:
See #205 for the original pull request.

The remaining unresolved feedback is primarily disagreements between me and @sonicfind as to how certain things should be done, which I'm hoping to re-discuss and resolve. I'd also appreciate some additional opinions on whatever points may remain, so that it's not just contention between us two and we can get this moved along lol


sonicfind (changes requested):

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.ProcessLists.cs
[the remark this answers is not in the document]
sonicfind: I don't think you're understanding here. ONLY increment in the else case. Don't just increment after every iteration and then decrement in the if/if-else cases.

Of course, you'll need to do the index overwrite *before* the incrementation, but in doing so, the other cases won't need `continue`.

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.ProcessLists.cs
```diff
@@ -211,9 +211,10 @@ private static void ConvertSoloEvents(ref NoteProcessParams noteProcessParams)
 
             // Manually move over events so we can avoid excessive element moves
             // from using RemoveAt or RemoveAll
-            int moveOffset = 0;
+            int insertIndex = 0;
 
-            for (int i = 0; i < chart.events.Count; ++i)
+            int i;
+            for (i = 0; i < chart.events.Count; i++, insertIndex++)
```
sonicfind, line 217: Why hold onto `i`? It's not useful outside of the loop (see next comment)

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.ProcessLists.cs
```diff
@@ -253,23 +254,18 @@ private static void ConvertSoloEvents(ref NoteProcessParams noteProcessParams)
...
+                chart.events[insertIndex] = ev;
             }
 
             // Trim off remaining excess elements from any removals
-            if (moveOffset > 0)
-            {
-                chart.events.RemoveRange(chart.events.Count - moveOffset, moveOffset);
-            }
+            int removed = i - insertIndex;
+            chart.events.RemoveRange(chart.events.Count - removed, removed);
```
sonicfind, line 268: `chart.events.Count - removed` = `insertIndex`, so that calculation is unnecessary; and `i` = `chart.events.Count`, which means you can just use `Count` instead for consistency.

sonicfind (changes requested):

TheNathannator:
really feel like we're just getting into the weeds of pedantry now lol

YARG.Core/MoonscraperChartParser/IO/Chart/ChartReader.ProcessLists.cs
TheNathannator: I had that before but it wasn't working, due to a (now unrelated) issue I was having previously which necessitated the decrements in the loop.

PR #235, opened 2025-02-05
Improve Pro Keys bot

wyrdough:
This PR modifies the pro keys bot in two ways.

Firstly, a bug that causes extended sustains to be dropped is fixed.

Secondly, the bot to is changed to release pressed keys 166ms after being pressed unless certain conditions are met.

The conditions that modify the key hold time are:
- Sustains are held for as long as they continue
- If the key is being played again it is released at half the time between the last press and next note
- ~~When the last note(s) have been played, any pressed keys are held for 1.6 seconds.~~

Note that a new field called _keyPressTimes is added to YargProKeysEngine. This may seem duplicative of the existing KeyPressTimes on ProKeysEngine, but that one is unsuitable for this purpose because it is updated on both press and release and is sometimes modified by hit logic.

As for the why, the reason for the first change is obvious. The second is because it makes me happy to have a bot that looks somewhat more natural to play with. Also, it will make future setlist wave preview videos look nicer.

### DIFF
--- a/YARG.Core/Engine/ProKeys/Engines/YargProKeysEngine.cs
+++ b/YARG.Core/Engine/ProKeys/Engines/YargProKeysEngine.cs
@@ -327,41 +327,58 @@ namespace YARG.Core.Engine.ProKeys.Engines
 
             var note = Notes[NoteIndex];
 
-            // Release keys that are not associated with an active sustain
+            // Release no longer needed keys
             int key = 0;
             for (var mask = KeyMask; mask > 0; mask >>= 1)
             {
                 // If it is a sustain, we do not drop
                 bool keySustained = false;
-                // Don't drop keys if the last note was within 166ms
+                // Don't drop keys if the note isn't over 166ms stale and no more than four have been played since
                 bool keyTooRecent = false;
                 // We do drop the key if the note we're processing is on this key
                 bool currentKey = false;
 
-                for (var i = 0; i < ActiveSustains.Count; i++)
+                foreach(var sustain in ActiveSustains)
                 {
-                    if (ActiveSustains[i].Note.Key == key)
+                    if (sustain.Note.Key == key)
                     {
                         keySustained = true;
                         break;
                     }
                 }
 
-                // C# does shortcut, right?
-                keyTooRecent = NoteIndex > 0 && Notes[NoteIndex - 1].Time > CurrentTime - 0.166f;
-
-                // TODO: Do child notes need to be checked here?
-                currentKey = time >= note.Time && note.Key == key;
-
-                if ((mask & 1) == 1 && ((!keySustained && !keyTooRecent) || currentKey))
+                if (!keySustained)
                 {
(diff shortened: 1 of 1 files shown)


TheNathannator:
I tested this on Du Hast from RB3 and there are numerous occasions there where the bot still drops extended sustains. I *can* confirm though that it handles fast notes and glissandos well, those look fine with these changes.

wyrdough:
> I tested this on Du Hast from RB3 and there are numerous occasions there where the bot still drops extended sustains. I _can_ confirm though that it handles fast notes and glissandos well, those look fine with these changes.

Ironically, the original version didn't do that. It got somewhat broken again because of the logic that drops all notes in a chord at the same time. It's fixed now.

The fix did require allocating an extra array of bools in UpdateBot, but in exchange ActiveSustains only has to be evaluated once per bot update instead of it being checked for every key in every update.

YARG.Core/Engine/ProKeys/Engines/YargProKeysEngine.cs
```diff
@@ -320,31 +330,111 @@ protected override bool CanNoteBeHit(ProKeysNote note)
 
         protected override void UpdateBot(double time)
         {
-            if (!IsBot || NoteIndex >= Notes.Count)
+            float        botNoteHoldTime = 0.166f;
+            ProKeysNote? note = null;
+            bool[]       keysInSustain = new bool[(int) ProKeysAction.Key25 + 1];
```
TheNathannator, line 335: This is the worst place to be allocating an array lol, you want to avoid allocations in code that is called per-frame to avoid repeated garbage collection stutters.

Use a bitmask instead (`int` is 32 bits wide, so it can handle all 25 keys):

```suggestion
            int keysInSustain = 0;
```

Using OR to mark a key as set:

```cs
            foreach (var sustain in ActiveSustains)
            {
                keysInSustain |= 1 << sustain.Note.Key;
            }
```

And AND to later test whether it is set:
```cs
                    if (note is not null && (keysInSustain & (1 << key)) != 0 && time >= note.Time)
                    {
                        keyProtected = false;
                    }
```

Or at least a stack-allocated `Span<bool>` instead:
```suggestion
            Span<bool> keysInSustain = stackalloc bool[(int) ProKeysAction.Key25 + 1];
```

Or, even more minimum, a cached array which is reset on every update:
```diff
         private KeyPressedTimes[] _keyPressedTimes = new KeyPressedTimes[(int)ProKeysAction.Key25 + 1];
+        private bool[] _keysInSustain = new bool[(int)ProKeysAction.Key25 + 1];
```

You never want to make temporary allocations in code that gets run frequently

TheNathannator (changes requested):

YARG.Core/Engine/ProKeys/Engines/YargProKeysEngine.cs
wyrdough: Ok, keysInSustain is a bitmask now. Thanks for the feedback and the learning. :)

PR #237, opened 2025-02-07
Replace Native STBImage Plugins With Custom-Tuned STBImageSharp

sonicfind:
Public Domain, so no license required


TheNathannator:
Even though no attribution is required, I'd say we should add a source + mention in the readme under [External Licenses](https://github.com/YARC-Official/YARG.Core?tab=readme-ov-file#-external-licenses) anyways so this can be tracked to the original source more easily.

sonicfind:
good call

PR #239, opened 2025-02-12
Remove underscores from LightingLookup

Pantotone:
Trying to fix the [error that appeared on recent nightly build](https://github.com/YARC-Official/YARG-BleedingEdge/actions/runs/13285521832/job/37093291163#step:17:27 "https://github.com/YARC-Official/YARG-BleedingEdge/actions/runs/13285521832/job/37093291163#step:17:27") about the underscores removed from the LightingTypes enum on #236 but not on the LightingLookup

Not sure if this is correct, needs a review from @TheFatBastid 

Also created a [build test](https://github.com/YARC-Official/YARG-BleedingEdge/actions/runs/13293487667)


EliteAsian123:
Thank you, it appears this was definitely the intention, however, it seems like that PR was not done correctly. I should've probably taken a closer look instead of just merging. This definitely appears to fix the issue.

PR #240, opened 2025-02-27
Add support for Ultrastar songs

ffio1:
This draft PR adds support for importing Ultrastar vocal tracks (either solo, or duet) to become playable in YARG.  I found this on the backlog and thought it would be a cool to add. https://yarg.youtrack.cloud/issue/YARG-78/Ultrastar-format-support

The reason I have this as a draft PR is to mainly get feedback on the overall layout of what I added, especially in regards to caching. The caching logic was mostly copy-paste from INI, but separate, so I may have missed something. The parsing logic is fairly simple as the Ultrastar format is just text, though with the combination of metadata and chart in one file, it resulted in having to parse the text file multiple times to get the specific data that we want out of it (one pass for metadata, another for the available players / preview time, and another for the chart). This can be optimized into one pass if needed.

Additionally, a question for how to handle preview start time. The documentation for Ultrastar Play says, "Defaults to halfway through the song." (https://github.com/UltraStar-Deluxe/Play/wiki/UltraStar-txt-File-Format) but that is not what is done in reality. What is done inside the game is that it takes every vocal note, puts them all in a list sorted by their beat, gets the note that has the index 1/3rd of the way in, gets the beat of said note, and the preview starts at that time ☨( https://github.com/UltraStar-Deluxe/Play/blob/master/UltraStar%20Play/Assets/Common/SongPreviewControl.cs#L154 ). That is how it is currently implemented in Ultrastar Play. Is this the behavior we want in YARG? I would say yes for consistency, but it also is an edge case that we may not want to deal with.

☨ - Ultrastar Play is on a build of Unity that has an issue that makes the preview timing even worse as there is an issue when decoding MP3 files where the time reported by unity is not the actual time inside the MP3. I don't think we want to mimic this as it's an engine bug, but just wanted to point this out if anyone sees it during testing ( https://discussions.unity.com/t/audioclip-length-is-incorrect-when-loading-from-webrequest-getaudioclip/834321/22 | https://github.com/UltraStar-Deluxe/Play/issues/504 )

### DIFF
--- a/YARG.Core/IO/TextReader/YARGTextReader.cs
+++ b/YARG.Core/IO/TextReader/YARGTextReader.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Text;
 using YARG.Core.Extensions;
 
@@ -10,15 +11,18 @@ namespace YARG.Core.IO
         public static readonly TChar NEWLINE;
         public static readonly TChar OPEN_BRACKET;
         public static readonly TChar CLOSE_BRACE;
+        public static readonly TChar POUND_SIGN;
 
         static unsafe TextConstants()
         {
             int newline = '\n';
             int openBracket = '[';
             int closeBrace = '}';
+            int poundSign = '#';
             NEWLINE = *(TChar*) &newline;
             OPEN_BRACKET = *(TChar*) &openBracket;
             CLOSE_BRACE = *(TChar*) &closeBrace;
+            POUND_SIGN = *(TChar*) &poundSign;
         }
     }
 
@@ -227,14 +231,14 @@ namespace YARG.Core.IO
             }
         }
 
-        public static unsafe string ExtractModifierName<TChar>(ref YARGTextContainer<TChar> container)
+        public static unsafe string ExtractModifierName<TChar>(ref YARGTextContainer<TChar> container, int splitChar = '=')
             where TChar : unmanaged, IConvertible
         {
             int length = 0;
             while (container.Position + length < container.Length)
             {
                 int val = container[length];
-                if (val <= WHITESPACE_LIMIT || val == '=')
+                if (val <= WHITESPACE_LIMIT || val == splitChar)
                 {
                     break;
                 }
@@ -330,18 +334,27 @@ namespace YARG.Core.IO
             return Decode(container.GetBuffer() + stringBegin, stringEnd - stringBegin, ref container);
         }
 
-        public static bool ExtractBoolean<TChar>(in YARGTextContainer<TChar> text)
+        public static bool ExtractBoolean<TChar>(in YARGTextContainer<TChar> text, string trueString = "true")
             where TChar : unmanaged, IConvertible
         {
             return !text.IsAtEnd() && text.Get() switch
             {
                 '1' => true,
-                _ => text.Position + 4 <= text.Length &&
-                    (text[0] | CharacterExtensions.ASCII_LOWERCASE_FLAG) is 't' &&
-                    (text[1] | CharacterExtensions.ASCII_LOWERCASE_FLAG) is 'r' &&
-                    (text[2] | CharacterExtensions.ASCII_LOWERCASE_FLAG) is 'u' &&
-                    (text[3] | CharacterExtensions.ASCII_LOWERCASE_FLAG) is 'e',
+                _ => text.Position + trueString.Length <= text.Length && MatchTrueString(text, trueString)
             };
+        }
+
+        private static bool MatchTrueString<TChar>(in YARGTextContainer<TChar> text, string trueString)
+            where TChar : unmanaged, IConvertible
+        {
+            for (int i = 0; i < trueString.Length; i++)
+            {
+                if ((text[i] | CharacterExtensions.ASCII_LOWERCASE_FLAG) != char.ToLower(trueString[i]))
+                {
+                    return false;
+                }
+            }
+            return true;
         }
 
         public static bool TryExtract<TChar, TNumber>(ref YARGTextContainer<TChar> text, out TNumber value)
--- a/YARG.Core/MoonscraperChartParser/IO/Ultrastar/UltrastarReader.cs
+++ b/YARG.Core/MoonscraperChartParser/IO/Ultrastar/UltrastarReader.cs
@@ -94,14 +94,10 @@ namespace YARG.Core.MoonscraperChartParser.IO.Ultrastar
 
             ChartSettings chartSettings = new ChartSettings(0, 0, 0, 120f, false, false);
             var song = new MoonSong(DEFAULT_RESOLUTION);
-            var charts = GetChartsForPlayer(chartSettings, song);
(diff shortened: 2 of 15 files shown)


YARG.Core/IO/TextReader/YARGTextReader.cs
```diff
@@ -10,15 +11,18 @@ public static class TextConstants<TChar>
...
         public static readonly TChar OPEN_BRACKET;
         public static readonly TChar CLOSE_BRACE;
+        public static readonly TChar POUND_SIGN;
 
         static unsafe TextConstants()
         {
             int newline = '\n';
             int openBracket = '[';
             int closeBrace = '}';
+            int poundSign = '#';
```
santorfo, line 21: poundSign could be mistaken for '£' (the British Pound currency sign). The unicode consortium calls it the [Number Sign](https://en.wikipedia.org/wiki/List_of_Unicode_characters#Latin_script).

YARG.Core/Song/Entries/Ultrastar/SongEntry.Ultrastar.cs
```diff
@@ -0,0 +1,522 @@
...
+                return ScanResult.NoName;
+            }
+
+            if (modifiers.Extract("version", out string versionString))
+            {
+                entry._ultrastarVersion = SongUltrastarHandler.ConvertVersionToEnum(versionString);
+            }
+            else
+            {
+                entry._ultrastarVersion = UltrastarVersion.V1_0_0;
```
santorfo, line 307: In the [UltraStar File Format Spec](https://github.com/UltraStar-Deluxe/format/blob/main/spec.md#33-the-version-header), it's mentioned that if no VERSION header is included, then it should default to Version 0.3.0

YARG.Core/Song/Entries/Ultrastar/SongEntry.Ultrastar.cs
```diff
@@ -0,0 +1,522 @@
...
+            // The preview would start at note 66, meaning the preview starts at 110s into the song audio.
+            if (noteBeats.Count > 0 && bpm != 0)
+            {
+                int noteIndex = (int)((noteBeats.Count - 1) * 0.33f);
+                int beatOfPreview = noteBeats[noteIndex];
+                double secondsPerBeat = 60 / bpm;
+                double timeOfPreviewSeconds = gapSeconds + (beatOfPreview * secondsPerBeat);
+
+                entry._metadata.Preview.Start = (long)(timeOfPreviewSeconds * SongMetadata.MILLISECOND_FACTOR);
+            }
```
santorfo, line 446: We should add a failsafe if this condition isn't met, otherwise entry._metadata.Preview.Start is never set

YARG.Core/MoonscraperChartParser/IO/Ultrastar/UltrastarReader.cs
```diff
@@ -0,0 +1,403 @@
...
+using YARG.Core.Utility;
+
+namespace YARG.Core.MoonscraperChartParser.IO.Ultrastar
+{
+    internal static partial class UltrastarReader
+    {
+        private const int MIDI_OFFSET_C4 = 60;
+
+        /// <summary>
+        /// BPM in Ultrastar file is 1/4th of the original, so multiplying to get correct bpm.
```
santorfo, line 22: This is only true for files below version 2.0.0.
https://github.com/UltraStar-Deluxe/format/blob/main/spec.md#34-the-bpm-header

YARG.Core/IO/TextReader/YARGTextReader.cs
ffio1: Thank you! I will update this. I was part of the generation that called this the hash / hashtag and/or the "pound sign", but I can change it to number sign to not confuse it with the British pound sign.

santorfo: Hash or Hashtag could work too, to be fair.

YARG.Core/Song/Entries/Ultrastar/SongEntry.Ultrastar.cs
ffio1: Thanks for sharing this! I will take a look over that spec, as I was originally using the spec from: https://github.com/UltraStar-Deluxe/Play/wiki/UltraStar-txt-File-Format as that was the one linked on the youtrack ( https://yarg.youtrack.cloud/issue/YARG-78/Ultrastar-format-support ) , and filling in the blanks based off their implementation.

I can change this to v0.3.0 as mentioned in that spec, but then anything that would've been played in Ultrastar Play that does not have a version specified would have a different version number specified in YARG. It shouldn't matter at the end of the day as I don't believe I'm doing anything specific with v1.0.0.

Song parsing:
https://github.com/UltraStar-Deluxe/Play/blob/361ec461ef96a8f084ff4001dfaaa19367f4d72d/UltraStar%20Play/Packages/playshared/Runtime/Songs/UltraStarSongFormat/UltraStarSongParser.cs#L90

Version enum:
https://github.com/UltraStar-Deluxe/Play/blob/361ec461ef96a8f084ff4001dfaaa19367f4d72d/UltraStar%20Play/Packages/playshared/Runtime/Songs/UltraStarSongFormat/UltraStarSongFormatVersion.cs#L21

In Ultrastar Play, when the TXT file is parsed it looks for the version header. If the header is not present, the enum's constructor will end up returning a value of v1.0.0, hence why I was returning v1.0.0 here.

YARG.Core/MoonscraperChartParser/IO/Ultrastar/UltrastarReader.cs
ffio1: Agreed, but is mentioned on that spec doc that the removal of said quadrupling has not been finalized yet, so I could make it so this only is true below version 2.0.0, but if they change the spec then we'd have to change our implementation again.

YARG.Core/Song/Entries/Ultrastar/SongEntry.Ultrastar.cs
ffio1: +1 to rejecting as I think we have a error of NoNotes that could be returned here.

YARG.Core/MoonscraperChartParser/IO/Ultrastar/UltrastarReader.cs
santorfo: Another thing to consider is what they wrote [here](https://usdx.eu/format/#specs)
"Speed of the song in Ultrastar - Beats Per Minute. Which is not the real song BPM. UltraStar BPM is quarter notes, thus it’s quadruple of the song’s BPM (ideally). "Ideally" means that any song could be tapped with any BPM (not even remotely related to the song’s BPM)."

We need to figure out if this could cause us any issues elsewhere. If the bpm value is only being used for ultrastar specific tasks, then we can treat it like USDX would, just have to make sure that we're not using a possibly arbitrary bpm value for calculations that would require the actual bpm of the song.

PR #242, opened 2025-03-12
Expose end_event tag in song metadata, allow getting of last note time from SongChart.

wyrdough:
To make it possible to be smarter about the delay in showing the score screen at the end of a song, it is necessary for GameManager to be able to find out whether the song has an end_event tag and get the end time of the last note in the chart.

At present, the end_event tag is not surfaced and SongChart.GetEndTime can only get whichever of the notes, phrases, or text events, making it impossible to determine the correct delay time.

This PR exposes these properties and adds support for getting the mixer audio level.

For the feature requiring this PR, see [YARG PR #993](https://github.com/YARC-Official/YARG/pull/993)

### DIFF
--- a/YARG.Core/Chart/SongChart.cs
+++ b/YARG.Core/Chart/SongChart.cs
@@ -329,12 +329,12 @@ namespace YARG.Core.Chart
             return totalStartTime;
         }
 
-        public double GetEndTime()
+        public double GetEndTime(bool noteOnly = false)
         {
-            static double TrackMax<TNote>(IEnumerable<InstrumentTrack<TNote>> tracks) where TNote : Note<TNote>
-                => tracks.Max((track) => track.GetEndTime());
-            static double VoxMax(IEnumerable<VocalsTrack> tracks)
-                => tracks.Max((track) => track.GetEndTime());
+            double TrackMax<TNote>(IEnumerable<InstrumentTrack<TNote>> tracks) where TNote : Note<TNote>
+                => tracks.Max((track) => track.GetEndTime(noteOnly));
+            double VoxMax(IEnumerable<VocalsTrack> tracks)
+                => tracks.Max((track) => track.GetEndTime(noteOnly));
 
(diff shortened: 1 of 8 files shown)


YARG.Core/Chart/SongChart.cs
```diff
@@ -362,6 +362,8 @@ static double VoxMax(IEnumerable<VocalsTrack> tracks)
             return totalEndTime;
         }
 
+        public double GetLastNoteEndTime() => GetEndTime(true);
```
TheNathannator, line 365: I'd prefer for this to be fully separated into a distinct function instead of shoving an optional parameter into `GetEndTime` just to re-use a bit of code here and there.

TheNathannator (changes requested):

wyrdough:
Ok, this should be right for real now, sorry about the commit spam

YARG.Core/Chart/Tracks/Vocals/VocalsPart.cs
```diff
@@ -61,6 +61,18 @@ public double GetEndTime()
...
+        public double GetLastNoteEndTime()
+        {
+            if (NotePhrases.Count > 0)
+            {
+                return NotePhrases[^1].TimeEnd;
+            }
+
+            // Not sure if we should be returning 0 or MaxValue or something else.
+            // 0 should be fine given what this is used for, but it feels somehow wrong
+            return 0;
```
TheNathannator, line 73: Very small nitpick: would this not be equivalent to just doing `GetEndTime` on `NotePhrases`?

...though now that I say it, I notice the same superfluous check for `Count > 0` is in the other start/end functions here. Something to investigate later I suppose lol

TheNathannator (approved):
Just a small nitpick I wanted to leave here, otherwise no more complaints!

Retargeting to `engine-fixes` to ensure that's easier to merge later, I'll handle any conflicts from that.

PR #245, opened 2025-03-25
Harmonies: Track notes that cross over phrase boundaries

thomeval:
This PR fixes a long standing issue with harmony charts that have notes that cross over phrase boundaries, such as this one:

image

This situation will cause two bugs to occur. Firstly, the full length of the affected note is counted as part of the left phrase. Depending on how much of that note happens to be in the next phrase, this can cause the left phrase to be more difficult, or even completely impossible, to obtain an AWESOME rating. Secondly, during the following phrase, the note is ignored entirely and cannot be hit at all. If this is the only note in that phrase, then the phrase is essentially considered "empty", and won't award any points.

Solves issue YARG-104. This PR also fixes an unrelated issue where the total number of SP phrases in a chart would be incorrect on the result screen for harmony players. Note that due to this PR, some harmony charts will have more valid phrases in them than before, raising the maximum possible score for them.

Screenshot 2025-03-25 115145

### DIFF
--- a/YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Vocals.cs
+++ b/YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Vocals.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using MoonscraperChartEditor.Song;
@@ -61,6 +61,7 @@ namespace YARG.Core.Chart
             var notePhrases = GetVocalsPhrases(moonChart, harmonyPart);
             var otherPhrases = GetPhrases(moonChart);
             var textEvents = GetTextEvents(moonChart);
+            TrimOrphanPhrases(notePhrases, otherPhrases);
             return new(isHarmony, notePhrases, otherPhrases, textEvents);
         }
 
@@ -80,6 +81,7 @@ namespace YARG.Core.Chart
             int moonNoteIndex = 0;
             int moonTextIndex = 0;
 
+            MoonNote? carriedNote = null;
             for (int moonPhraseIndex = 0; moonPhraseIndex < moonChart.specialPhrases.Count;)
             {
                 var moonPhrase = moonChart.specialPhrases[moonPhraseIndex++];
@@ -98,6 +100,11 @@ namespace YARG.Core.Chart
 
                     phraseTracker[moonPhrase2.type] = moonPhrase2;
                     moonPhraseIndex++;
+                }
+
+                if (carriedNote != null && carriedNote.tick + carriedNote.length < moonPhrase.tick)
+                {
+                    carriedNote = null;
                 }
 
                 // Go through each note and lyric in the phrase
@@ -150,11 +157,16 @@ namespace YARG.Core.Chart
                         continue;
                     }
 
+                    if (moonNote.tick + moonNote.length > moonPhrase.tick + moonPhrase.length)
+                    {
+                        carriedNote = moonNote;
+                    }
+
                     notes.Add(note);
                     previousNote = note;
                 }
 
-                if (notes.Count < 1)
+                if (notes.Count < 1 && carriedNote == null)
                 {
                     // This can occur on harmonies, HARM1 must contain phrases for all harmony parts
                     // so, for example, phrases with only HARM2/3 notes will cause this
@@ -167,6 +179,18 @@ namespace YARG.Core.Chart
 
             phrases.TrimExcess();
             return phrases;
+        }
+
+        private void TrimOrphanPhrases(List<VocalsPhrase> vocalPhrases, List<Phrase> otherPhrases)
+        {
+            for (int i = 0; i < otherPhrases.Count; i++)
+            {
+                if (!vocalPhrases.Any(e => e.Tick == otherPhrases[i].Tick))
+                {
+                    otherPhrases.RemoveAt(i);
+                    i--;
+                }
+            }
         }
 
         private void ProcessLyric(List<LyricEvent> lyrics, ReadOnlySpan<char> lyric, uint lyricTick,
--- a/YARG.Core/Engine/Vocals/VocalsEngine.cs
+++ b/YARG.Core/Engine/Vocals/VocalsEngine.cs
@@ -11,6 +11,8 @@ namespace YARG.Core.Engine.Vocals
     {
         protected const int POINTS_PER_PERCUSSION = 100;
 
+        protected VocalNote? CarriedVocalNote;
+
         public delegate void TargetNoteChangeEvent(VocalNote targetNote);
 
         public delegate void PhraseHitEvent(double hitPercentAfterParams, bool fullPoints);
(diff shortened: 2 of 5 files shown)


YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Vocals.cs
```diff
@@ -169,6 +181,18 @@ private List<VocalsPhrase> GetVocalsPhrases(MoonChart moonChart, int harmonyPart
             return phrases;
         }
 
+        private void TrimOrphanPhrases(List<VocalsPhrase> vocalPhrases, List<Phrase> otherPhrases)
+        {
+            for (int i = 0; i < otherPhrases.Count; i++)
+            {
+                if (!vocalPhrases.Any(e => e.Tick == otherPhrases[i].Tick))
```
TheNathannator, line 188: Going over the entire list of vocals phrases for every other phrase in the chart will scale very poorly for large vocals charts (as it is an O(n^2) operation). Instead, keep an index for what the current vocals phrase is, and increment that index whenever the current other phrase is outside the bounds of that vocals phrase (see the phrase handling in `GetVocalsPhrases` for reference).

As another note, try to avoid using LINQ if you can help it, it has both allocation and performance overhead which can cause additional load time issues here in the parsing stage.

YARG.Core/Engine/Vocals/VocalsEngine.cs
```diff
@@ -224,27 +226,32 @@ protected void MissNote(VocalNote note, double hitPercent)
         /// <returns>
         /// Gets the amount of ticks in the phrase.
         /// </returns>
-        protected static uint GetTicksInPhrase(VocalNote phrase)
+        protected uint GetTicksInPhrase(VocalNote phrase)
         {
             uint totalTime = 0;
-            foreach (var phraseNote in phrase.ChildNotes)
+            foreach (var noteInPhrase in phrase.ChildNotes.Where(e => !e.IsPercussion))
```
TheNathannator, line 232: *Especially* avoid using LINQ in engine code. This will be run repeatedly every frame, and add lots of allocation and performance overhead, which will result in GC pauses and stuttering. The `IsPercussion` check is placed directly in the loop for a reason.

YARG.Core/Engine/Vocals/VocalsEngine.cs
```diff
@@ -315,9 +322,22 @@ protected override void UpdateMultiplier()
...
+        protected virtual void UpdateCarriedNote(VocalNote phrase)
+        {
+            if (CarriedVocalNote != null && CarriedVocalNote.TotalTickEnd > phrase.TickEnd)
+            {
+                // Keep the current Carried Vocal Note if it continues past the end of the current phrase.
+                // This can happen for notes that span across 3 or more phrases.
+                return;
+            }
+
+            CarriedVocalNote = phrase.ChildNotes.FirstOrDefault(e => !e.IsPercussion && e.TotalTickEnd > phrase.TickEnd);
```
TheNathannator, line 339: Again avoid using LINQ in engine code that runs repeatedly. This should be unfolded into an explicit loop:

```suggestion
            CarriedVocalNote = null;
            foreach (var note in phrase.ChildNotes)
            {
                if (e.IsPercussion && e.TotalTickEnd > phrase.TickEnd)
                {
                    CarriedVocalNote = note;
                    break;
                }
            }
```

YARG.Core/Engine/Vocals/VocalsStats.cs
```diff
@@ -22,6 +23,8 @@ public class VocalsStats : BaseStats
         /// </summary>
         public uint TotalTicks => TicksHit + TicksMissed;
 
+        public bool HasCarryNote;
```
TheNathannator, line 26: I don't see this stat referenced anywhere, is it needed?

TheNathannator (changes requested):
It's not immediately obvious to me what approach this PR takes to fix the issue, can you provide more details?

In general though I think this solution might be more complex than it needs to be, in theory we can just split the crossing notes into two at the point where they cross phrases. (May also want to be wary of notes that cross non-contiguous phrases, not 100% certain what to do there.)

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Vocals.cs
```diff
@@ -150,11 +157,16 @@ private List<VocalsPhrase> GetVocalsPhrases(MoonChart moonChart, int harmonyPart
...
+                    {
+                        carriedNote = moonNote;
+                    }
+
                     notes.Add(note);
                     previousNote = note;
                 }
 
-                if (notes.Count < 1)
+                if (notes.Count < 1 && carriedNote == null)
```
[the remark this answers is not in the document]
thomeval: This handles cases where the carry note is the _only_ note in the phrase. In that case, we still need to add a phrase there so that players can actually hit the carry note. You can see an example of this in the screenshot in the description.

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Vocals.cs
```diff
@@ -80,6 +81,7 @@ private List<VocalsPhrase> GetVocalsPhrases(MoonChart moonChart, int harmonyPart
             int moonNoteIndex = 0;
             int moonTextIndex = 0;
 
+            MoonNote? carriedNote = null;
```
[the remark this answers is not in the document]
thomeval: Correct. We need to account for it both in this phrase, and the next.

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Vocals.cs
```diff
@@ -169,6 +181,18 @@ private List<VocalsPhrase> GetVocalsPhrases(MoonChart moonChart, int harmonyPart
             return phrases;
         }
 
+        private void TrimOrphanPhrases(List<VocalsPhrase> vocalPhrases, List<Phrase> otherPhrases)
```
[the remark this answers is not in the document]
thomeval: Yes, this is used to remove Star Power phrases that don't have a corresponding vocals phrase. This fixes the issue where the game incorrectly counts the total number of phrases in a HARM2/HARM3 chart as you can see from the screenshot.
As for your suggestion, is there any reason why we would want to have a "special" phrase that isn't in a vocals phrase? Even percussion sections are required to be in a vocals phrase, so it might make sense to enforce a rule that any significant part of a vocals chart must be inside a vocals phrase.

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Vocals.cs
thomeval: Should be resolved now.

YARG.Core/Engine/Vocals/VocalsEngine.cs
thomeval: Resolved

wyrdough:
It doesn't obviously break anything, so screw it, we'll test in production ;)

PR #246, opened 2025-03-28
Generate weak beatlines from sync track MIDI note 14

wyrdough:
Per charting team request, this PR adds support for specifying weak beats using MIDI note 14.

Before:
weak beatline off comparison

After:
weak beatline on comparison

### DIFF
--- a/YARG.Core/MoonscraperChartParser/IO/Midi/MidIOHelper.cs
+++ b/YARG.Core/MoonscraperChartParser/IO/Midi/MidIOHelper.cs
@@ -110,8 +110,9 @@ namespace MoonscraperChartEditor.Song.IO
         public const byte PRO_GUITAR_CHANNEL_PINCH_HARMONIC = 6;
 
         // Beat track notes
-        public const byte BEAT_STRONG = 12;
-        public const byte BEAT_WEAK = 13;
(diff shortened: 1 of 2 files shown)


YARG.Core/MoonscraperChartParser/IO/Midi/MidIOHelper.cs
```diff
@@ -112,6 +112,7 @@ internal static class MidIOHelper
         // Beat track notes
         public const byte BEAT_STRONG = 12;
         public const byte BEAT_WEAK = 13;
+        public const byte BEAT_SUPER_WEAK = 14;
```
TheNathannator, line 115: Would prefer for these constants to be renamed to `BEAT_MEASURE`, `BEAT_STRONG`, and `BEAT_WEAK` to match the `BeatlineType` enum, for clarity.

TheNathannator (changes requested):
Just a small naming nitpick.

YARG.Core/MoonscraperChartParser/IO/Midi/MidIOHelper.cs
wyrdough: Commit amended...

TheNathannator (approved):

PR #247, opened 2025-04-04
Add modifier to shift all notes into starting range, etc.

wyrdough:
This PR adds a new modifier to compress all ranges into the song's starting range. (assuming that the song supports range shifts)

For example, if a song starts with the GRY range and later shifts to YBO, the YBO notes will be transposed to GRY notes.

For more background on what range shifts even are, see the [relevant proposal](https://discord.com/channels/1086048856678084609/1344362559347621960) on Discord.

Currently a draft because we don't yet have a modifier icon.

### DIFF
--- a/YARG.Core/Chart/Events/ChartEventExtensions.cs
+++ b/YARG.Core/Chart/Events/ChartEventExtensions.cs
@@ -154,41 +154,6 @@ namespace YARG.Core.Chart
             }
         }
 
-        /// <summary>
-        /// Creates a new note group not including the indicated child
-        /// </summary>
-        /// <param name="notes"></param>
-        /// <param name="index"></param>
-        /// <param name="childIndex"></param>
-        public static void RemoveChildFromNote(this List<GuitarNote> notes, int index, int childIndex)
-        {
-            var oldNote = notes[index];
-            var newNote = notes[index].CloneWithoutChildNotes();
-
-            for(int i = 0; i < oldNote.ChildNotes.Count; i++)
-            {
-                if (childIndex != i)
-                {
-                    newNote.AddChildNote(oldNote.ChildNotes[i]);
-                }
-            }
-
-            // Stitch references
-            if (oldNote.PreviousNote != null)
-            {
-                newNote.PreviousNote = oldNote.PreviousNote;
-                oldNote.PreviousNote.NextNote = newNote;
-            }
-
-            if (oldNote.NextNote != null)
-            {
-                newNote.NextNote = oldNote.NextNote;
-                oldNote.NextNote.PreviousNote = newNote;
-            }
-
-            notes[index] = newNote;
-        }
-
         public static double GetStartTime<TEvent>(this List<TEvent> events)
             where TEvent : ChartEvent
         {
--- a/YARG.Core/Chart/Events/ChartEventExtensions.cs
+++ b/YARG.Core/Chart/Events/ChartEventExtensions.cs
@@ -118,58 +118,36 @@ namespace YARG.Core.Chart
 
                 notes.RemoveAt(index);
 
-                // If there is now only one note, previous and next both need to be set to null
-                if (notes.Count == 1)
-                {
-                    notes[0].PreviousNote = null;
-                    notes[0].NextNote = null;
-                    return;
-                }
-
-                // Past this point, we can be assured that notes[0] and notes[^1] are valid and don't refer
-                // to the same note
-
-                // Fix Previous/Next references
-                if (index == 0)
-                {
-                    // We were the first, so new 0's previous becomes null and everything else is fine
-                    notes[0].PreviousNote = null;
-                    return;
-                }
-
-                // We have already removed a note so index will equal notes.Count if it is referring
-                // to what was previously the last note in the chart
-                if (index == notes.Count)
-                {
-                    // We were the last, so new last's next becomes null and everything else is fine
-                    notes[^1].NextNote = null;
-                    return;
-                }
-
-                // We're somewhere in the middle, so we have to stitch references back together
-                // note is no longer good, so back to indexed access
-                notes[index - 1].NextNote = notes[index];
-                notes[index].PreviousNote = notes[index].PreviousNote;
-            }
-            else
-            {
-                // Promote first child to parent, reparent other children, and stitch references
-                var newParent = notes[index].ChildNotes[0].Clone();
-                for (var i = 1; i < notes[index].ChildNotes.Count; i++)
-                {
-                    newParent.AddChildNote(notes[index].ChildNotes[i]);
-                }
-
                 if (note.PreviousNote != null)
                 {
-                    newParent.PreviousNote = note.PreviousNote;
-                    note.PreviousNote.NextNote = newParent;
+                    note.PreviousNote.NextNote = note.NextNote;
                 }
 
                 if (note.NextNote != null)
                 {
-                    newParent.NextNote = note.NextNote;
-                    note.NextNote.PreviousNote = newParent;
+                    note.NextNote.PreviousNote = note.PreviousNote;
+                }
+            }
+            else
+            {
+                // Promote first child to parent, reparent other children, and stitch references
+                var newParent = note.ChildNotes[0].Clone();
+                for (var i = 1; i < note.ChildNotes.Count; i++)
+                {
+                    newParent.AddChildNote(note.ChildNotes[i]);
+                }
+
+                newParent.PreviousNote = note.PreviousNote;
+                newParent.NextNote = note.NextNote;
+
+                if (newParent.PreviousNote != null)
+                {
+                    newParent.PreviousNote.NextNote = newParent;
+                }
+
+                if (newParent.NextNote != null)
+                {
+                    newParent.NextNote.PreviousNote = newParent;
                 }
 
                 notes[index] = newParent;
--- a/YARG.Core/Chart/Events/ChartEventExtensions.cs
+++ b/YARG.Core/Chart/Events/ChartEventExtensions.cs
@@ -45,6 +45,8 @@ namespace YARG.Core.Chart
         // Remove note from list in place, adjusting references as required
         public static void RemoveNoteAt(this List<GuitarNote> notes, int index)
         {
+            var note = notes[index];
+
             // Check for invalid index
             if (index < 0 || index >= notes.Count)
             {
@@ -52,40 +54,32 @@ namespace YARG.Core.Chart
                 return;
             }
 
-            // If there's only one note with no children, we can just remove it and return, as there is
-            // nowhere left to move the flags and nothing to stitch back together
-            if (notes.Count == 1 && notes[index].ChildNotes.Count == 0)
-            {
-                notes.RemoveAt(index);
-                return;
-            }
-
-            if (notes[index].ChildNotes.Count == 0)
+            if (note.ChildNotes.Count == 0)
             {
                 // If there are no children, we have to deal with moving flags
-                if (notes[index].IsSoloStart && !notes[index].IsSoloEnd)
+                if (note.IsSoloStart && !note.IsSoloEnd)
                 {
                     // This is a single note that is a solo start, we have to move it to the
                     // NEXT note (we don't want to extend the solo).
-                    if (index + 1 < notes.Count)
+                    if (note.NextNote != null)
                     {
-                        notes[index + 1].Flags |= NoteFlags.SoloStart;
+                        note.NextNote.Flags |= NoteFlags.SoloStart;
                         // Also add it to the child notes
-                        foreach (var childNote in notes[index + 1].ChildNotes)
+                        foreach (var childNote in note.NextNote.ChildNotes)
                         {
                             childNote.Flags |= NoteFlags.SoloStart;
                         }
                     }
                 }
-                else if (!notes[index].IsSoloStart && notes[index].IsSoloEnd)
+                else if (!note.IsSoloStart && note.IsSoloEnd)
                 {
                     // This is a single note that is a solo end, we have to move it to the
                     // PREVIOUS note (we don't want to extend the solo).
-                    if (index > 0)
+                    if (note.PreviousNote != null)
                     {
-                        notes[index - 1].Flags |= NoteFlags.SoloEnd;
+                        note.PreviousNote.Flags |= NoteFlags.SoloEnd;
                         // Also add it to the child notes
-                        foreach (var childNote in notes[index - 1].ChildNotes)
+                        foreach (var childNote in note.PreviousNote.ChildNotes)
                         {
                             childNote.Flags |= NoteFlags.SoloEnd;
                         }
@@ -93,29 +87,29 @@ namespace YARG.Core.Chart
                 }
 
 
-                if (notes[index].IsStarPowerStart && !notes[index].IsStarPowerEnd)
+                if (note.IsStarPowerStart && !note.IsStarPowerEnd)
                 {
                     // Make the next note the SP start, so as to not extend the SP section
-                    if (index + 1 < notes.Count)
+                    if (note.NextNote != null)
                     {
-                        notes[index + 1].Flags |= NoteFlags.StarPowerStart;
+                        note.NextNote.Flags |= NoteFlags.StarPowerStart;
                         // Also add it to the child notes
-                        foreach (var childNote in notes[index + 1].ChildNotes)
+                        foreach (var childNote in note.NextNote.ChildNotes)
                         {
                             childNote.Flags |= NoteFlags.StarPowerStart;
                         }
                     }
                 }
 
-                if (!notes[index].IsStarPowerStart && notes[index].IsStarPowerEnd)
+                if (!note.IsStarPowerStart && note.IsStarPowerEnd)
                 {
                     // This is a single kick drum note that is a starpower end, we have to move it to the
                     // PREVIOUS note (we don't want to extend the starpower section).
-                    if (index > 0)
+                    if (note.PreviousNote != null)
                     {
-                        notes[index - 1].Flags |= NoteFlags.StarPowerEnd;
+                        note.PreviousNote.Flags |= NoteFlags.StarPowerEnd;
                         // Also add it to the child notes
-                        foreach (var childNote in notes[index - 1].ChildNotes)
+                        foreach (var childNote in note.PreviousNote.ChildNotes)
                         {
                             childNote.Flags |= NoteFlags.StarPowerEnd;
                         }
@@ -153,8 +147,9 @@ namespace YARG.Core.Chart
                 }
 
                 // We're somewhere in the middle, so we have to stitch references back together
+                // note is no longer good, so back to indexed access
                 notes[index - 1].NextNote = notes[index];
-                notes[index].PreviousNote = notes[index - 1];
+                notes[index].PreviousNote = notes[index].PreviousNote;
             }
             else
             {
@@ -165,16 +160,16 @@ namespace YARG.Core.Chart
                     newParent.AddChildNote(notes[index].ChildNotes[i]);
                 }
 
-                if (index > 0)
+                if (note.PreviousNote != null)
                 {
-                    newParent.PreviousNote = notes[index - 1];
-                    notes[index - 1].NextNote = newParent;
+                    newParent.PreviousNote = note.PreviousNote;
+                    note.PreviousNote.NextNote = newParent;
                 }
 
-                if (index + 1 < notes.Count)
+                if (note.NextNote != null)
                 {
-                    newParent.NextNote = notes[index + 1];
-                    notes[index + 1].PreviousNote = newParent;
+                    newParent.NextNote = note.NextNote;
+                    note.NextNote.PreviousNote = newParent;
                 }
 
                 notes[index] = newParent;
@@ -189,27 +184,28 @@ namespace YARG.Core.Chart
         /// <param name="childIndex"></param>
         public static void RemoveChildFromNote(this List<GuitarNote> notes, int index, int childIndex)
         {
+            var oldNote = notes[index];
             var newNote = notes[index].CloneWithoutChildNotes();
 
-            for(int i = 0; i < notes[index].ChildNotes.Count; i++)
+            for(int i = 0; i < oldNote.ChildNotes.Count; i++)
             {
                 if (childIndex != i)
                 {
-                    newNote.AddChildNote(notes[index].ChildNotes[i]);
+                    newNote.AddChildNote(oldNote.ChildNotes[i]);
                 }
             }
 
             // Stitch references
-            if (index > 0)
+            if (oldNote.PreviousNote != null)
             {
-                newNote.PreviousNote = notes[index - 1];
-                notes[index - 1].NextNote = newNote;
+                newNote.PreviousNote = oldNote.PreviousNote;
+                oldNote.PreviousNote.NextNote = newNote;
             }
 
-            if (index + 1 < notes.Count)
+            if (oldNote.NextNote != null)
             {
-                newNote.NextNote = notes[index + 1];
-                notes[index + 1].PreviousNote = newNote;
+                newNote.NextNote = oldNote.NextNote;
+                oldNote.NextNote.PreviousNote = newNote;
             }
 
             notes[index] = newNote;
--- a/YARG.Core/Chart/Events/ChartEventExtensions.cs
+++ b/YARG.Core/Chart/Events/ChartEventExtensions.cs
@@ -41,6 +41,171 @@ namespace YARG.Core.Chart
             return newNotes;
         }
 
+        // Remove note from list in place, adjusting references as required
+        public static void RemoveNoteAt(this List<GuitarNote> notes, int index)
+        {
+            if (notes[index].ChildNotes.Count == 0)
+            {
+                // The simple case, when there are no children
+                // Well, simple except for having to deal with all the flags that may need to move
+                if (notes[index].IsSoloStart && !notes[index].IsSoloEnd)
+                {
+                    // This is a single note that is a solo start, we have to move it to the
+                    // NEXT note (we don't want to extend the solo).
+                    if (index < notes.Count)
+                    {
+                        notes[index + 1].Flags |= NoteFlags.SoloStart;
+                        // Also add it to the child notes
+                        foreach (var childNote in notes[index].ChildNotes)
+                        {
+                            childNote.Flags |= NoteFlags.SoloStart;
+                        }
+                    }
+                }
+
+                if (notes[index].IsSoloEnd)
+                {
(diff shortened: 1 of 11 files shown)


YARG.Core/Chart/Events/ChartEventExtensions.cs
```diff
@@ -41,6 +41,171 @@ public static List<TNote> DuplicateNotes<TNote>(this List<TNote> notes)
...
+                {
+                    newParent.PreviousNote = notes[index - 1];
+                    notes[index - 1].NextNote = newParent;
+                }
+                else
+                {
+                    newParent.PreviousNote = null;
+                }
+
+                if (index < notes.Count - 1)
```
sonicfind, line 151: Prefer `index + 1` to match

YARG.Core/Chart/Events/ChartEventExtensions.cs
```diff
@@ -41,6 +41,171 @@ public static List<TNote> DuplicateNotes<TNote>(this List<TNote> notes)
...
+            var newNote = notes[index].CloneWithoutChildNotes();
+
+            for(int i = 0; i < notes[index].ChildNotes.Count; i++)
+            {
+                if (childIndex == i)
+                {
+                    continue;
+                }
+
+                newNote.AddChildNote(notes[index].ChildNotes[i]);
```
sonicfind, line 182: For single-liners, just put the operation in the branch.

YARG.Core/Chart/Events/ChartEventExtensions.cs
```diff
@@ -41,6 +41,171 @@ public static List<TNote> DuplicateNotes<TNote>(this List<TNote> notes)
...
+
+            // Stitch references
+            if (index > 0)
+            {
+                newNote.PreviousNote = notes[index - 1];
+                notes[index - 1].NextNote = newNote;
+            }
+            else
+            {
+                newNote.PreviousNote = null;
```
sonicfind, line 193: `PreviousNote` is always `null` on clone

YARG.Core/Chart/Events/ChartEventExtensions.cs
```diff
@@ -41,6 +41,171 @@ public static List<TNote> DuplicateNotes<TNote>(this List<TNote> notes)
...
+            }
+
+            if (index < notes.Count - 1)
+            {
+                newNote.NextNote = notes[index + 1];
+                notes[index + 1].PreviousNote = newNote;
+            }
+            else
+            {
+                newNote.NextNote = null;
```
sonicfind, line 203: `NextNote` is always `null` on clone

YARG.Core/Chart/Events/ChartEventExtensions.cs
```diff
@@ -41,6 +41,171 @@ public static List<TNote> DuplicateNotes<TNote>(this List<TNote> notes)
...
+            {
+                newNote.PreviousNote = notes[index - 1];
+                notes[index - 1].NextNote = newNote;
+            }
+            else
+            {
+                newNote.PreviousNote = null;
+            }
+
+            if (index < notes.Count - 1)
```
sonicfind, line 196: Prefer `index + 1` to match

YARG.Core/Chart/SongChart.AutoGeneration.cs
```diff
@@ -463,5 +463,153 @@ private static bool IsIdealDrumActivationNote(DrumNote note, Instrument instrume
...
+
+                    for (var i = 0; i < textEvents.Count; i++)
+                    {
+                        var time = textEvents[i].Time;
+                        var tick = textEvents[i].Tick;
+                        double timeEnd;
+                        uint tickEnd;
+
+                        // Overlaps will lead to bad things, so the end of one phrase is a tick before the beginning of the next
+                        if (i < textEvents.Count - 1)
```
sonicfind, line 494: Prefer `i + 1` to match

YARG.Core/Chart/SongChart.AutoGeneration.cs
```diff
@@ -463,5 +463,153 @@ private static bool IsIdealDrumActivationNote(DrumNote note, Instrument instrume
...
+                        {
+                            tickEnd = textEvents[i + 1].Tick - 1;
+                            timeEnd = SyncTrack.TickToTime(tickEnd);
+                        }
+                        else
+                        {
+                            // This is the last range shift, so it ends once all the notes are over
+                            tickEnd = instrumentDifficulty.Notes[^1].TickEnd;
+                            timeEnd = instrumentDifficulty.Notes[^1].TimeEnd;
```
sonicfind, line 503: Making the assumption that there are notes. Should there be? yes, but you have to account for that regardless.

sonicfind: Also, calling the indexer twice instead of holding onto the Note instance as a local.

YARG.Core/Chart/SongChart.AutoGeneration.cs
```diff
@@ -463,5 +463,153 @@ private static bool IsIdealDrumActivationNote(DrumNote note, Instrument instrume
...
+                if ((int) instrumentDifficulty.Difficulty != eventDifficulty)
+                {
+                    continue;
+                }
+
+                // We have a valid range shift, so place it in the appropriate difficulty bin
+                textEvents.Add(textEvent);
+            }
+            return textEvents;
+        }
```
sonicfind, line 613: If you're defining a function for the explicit purpose of extracting out events with well defined variables/values, return a list of structs specifically built to provide those values instead of just returning the unprocessed strings events again.

YARG.Core/Chart/SongChart.AutoGeneration.cs
```diff
@@ -463,5 +463,153 @@ private static bool IsIdealDrumActivationNote(DrumNote note, Instrument instrume
...
+                        // this point, so we can rely on splitEvent.Length being exactly 3 or 4 here.
+                        if (splitEvent.Length == 3 || !int.TryParse(splitEvent[3], out size))
+                        {
+                            size = instrumentDifficulty.Difficulty switch
+                            {
+                                Difficulty.Easy   => 3,
+                                Difficulty.Medium => 4,
+                                _                 => 5,
+                            };
+                        }
```
sonicfind, line 522: (See comment on the `ParseRangeShifts` function)
Case and point: double-splitting each string is a big no.

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +34,173 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+
+            var shifts = difficulty.RangeShiftEvents;
+
+            // We're shifting all ranges into the first range, so set rangeTo to the first range
+            int rangeTo = shifts[0].Range;
+            // These need to be initialized here since the first range won't have set them in the loop
+            int rangeFrom = shifts[0].Range;
+            int shiftIndex = 0;
+            int shiftAmount = rangeFrom - rangeTo;
+            GuitarNote note;
```
sonicfind, line 58: This being outside of the loop is kinda pointless when it's a reference type.

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +34,173 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+
+            for (var i = 0; i < difficulty.Notes.Count; i++)
+            {
+                note = difficulty.Notes[i];
+
+                if (note.Time >= shifts[shiftIndex].TimeEnd && shiftIndex + 1 < shifts.Count)
```
sonicfind, line 64: Flip the conditional to check the index first before doing the time comparison. No real reason to check the time if we're not going to run the operation anyways.

sonicfind: Also, compare against the *next* phrase's start time, NOT the *current* phrase's end time. Remember that the end tick is always one tick before the next phrase. So if you have a note that lands on that tick, it's technically no longer in *any* phrase at all. Sure it "works", but that still makes it ambiguous.

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +34,173 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+                if (rangeFrom == rangeTo)
+                {
+                    continue;
+                }
+
+                // Leave open notes as they are
+                if (note.Fret == (int) FiveFretGuitarFret.Open)
+                {
+                    continue;
+                }
```
sonicfind, line 80: Just combine these two conditions. It's not like you're throwing an exception here that needs explicit messaging.

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +34,173 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+                if (shiftAmount > 0)
+                {
+                    note.NoteMask >>= shiftAmount;
+                    note.DisjointMask >>= shiftAmount;
+                }
+                else
+                {
+                    note.NoteMask <<= shiftAmount;
+                    note.DisjointMask <<= shiftAmount;
+                }
```
sonicfind, line 93: No negation here, so it still fails

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
sonicfind: Also, this fails for open note chords. You'd still need to shift the upper notes down within range.

sonicfind: Plus, this makes the dangerous assumption that the open note is *always* the parent note. While most charts abide by this rule, none of the moonscraper parser code nor the moonsongloader code forces that relationship from what I can tell (through .chart at least).

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +34,173 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+                    if (shiftAmount > 0)
+                    {
+                        child.NoteMask >>= shiftAmount;
+                        child.DisjointMask >>= shiftAmount;
+                    }
+                    else
+                    {
+                        child.NoteMask <<= shiftAmount;
+                        child.DisjointMask <<= shiftAmount;
+                    }
```
sonicfind, line 110: No negation here, so it still fails

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +34,173 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+                if (!outOfRange)
+                {
+                    continue;
+                }
+
+                // Check if parent is out of range, replacing parent with a child if so, and keep doing
+                // that until we get a parent that is in range or we run out of children
+                while (note.Fret is < (int) FiveFretGuitarFret.Green or > (int) FiveFretGuitarFret.Orange)
+                {
+                    difficulty.Notes.RemoveNoteAt(i);
```
sonicfind, line 134: From an API perspective, this does not appropriately communicate the operations being performed by this function. When I look at this, without looking into the implementation details of the function itself, I think that it's straight removing that note index entirely, not just _potentially_ doing so. That's an inherent issue with readability/maintainability. 

But I can't be too picky here, as the entire note architecture is just bad.

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +34,173 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+                        {
+                            continue;
+                        }
+
+                        if (prevNote.Tick + prevNote.TickLength > child.Tick)
+                        {
+                            prevNote.TickLength = child.Tick - prevNote.Tick;
+                        }
+                    }
+                }
```
sonicfind, line 200: This fails as it's not just the directly previous note that may overlap. The note in question could be many many indices before this.

Which is... quite a big problem.

YARG.Core/Game/YargProfile.cs
```diff
@@ -93,6 +95,7 @@ public YargProfile()
             NoteSpeed = 6;
             HighwayLength = 1;
             LeftyFlip = false;
+            RangeEnabled = true;
```
sonicfind, line 98: Why is this defaulted to true?

wyrdough: Because it is intended that range shift markers be enabled by default for charts that have range shifts with the option to disable if the user so desires.

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
wyrdough: @TheNathannator said a while ago on Discord that there is a defined ordering, though it may be that in an open chord the open is always a child since the rightmost note on the track is always the parent note in my testing. (I have not actually tested open chords..didn't think of it since range shifts are only used on easy and medium charts)

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
wyrdough: Unless you're referring to the case where there is an open chord, I don't understand what you're saying...

YARG.Core/Game/YargProfile.cs
sonicfind: ohhhhh this is the shifting not the compression. gotcha

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
grishhung: Discussed offline; please note that the behavior of open notes during shifting should be that they just don't move. For example, a `P` + `-----O` open chord during a `--YBO` section on easy should turn into a `P` + `--Y--` chord once it's condensed down to a `GRY--` section with this modifier.

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
wyrdough: Oh, what you're saying is that this will end up shifting left by a negative index when the index needs to be positive *slaps forehead*

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
wyrdough: It does always remove a note, but it is true that the index may not be removed in the case where the note at that index has a child note that is then promoted.

If you have a suggestion for a better name, I'm happy to take suggestions. :)

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -171,30 +187,58 @@ public static void CompressGuitarRange(this InstrumentDifficulty<GuitarNote> dif
...
 
-                // Cut off sustains that would overlap
-                foreach (var prevNote in difficulty.Notes[i-1].AllNotes)
+                // For each note in our note group, check to see if the identified previous note is a sustain
+                foreach (var parentOrChild in note.AllNotes)
                 {
-                    if (!prevNote.IsSustain)
+                    // TODO: How the fark is Fret ever 0 here? That's out of range and we should have discarded all out
+                    // of range frets earlier!
+                    if (overlapIndexes[parentOrChild.Fret - 1] == -1)
```
wyrdough, line 224: Need some help here. Not sure if it's just because my brain is fried, but I can't for the life of me figure out how a note that has Fret = 0 is still around at this point. They should have been discarded earlier in the process.

wyrdough: Figured it out after staring at it a while. Calling RemoveChildFromNote updates the instrumentDifficulty.Notes list, but the local variable note was holding a reference to the old version of the note that still had the removed child attached.

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
wyrdough: Updated this to look back as far as necessary to find the most recent note in the lane.

YARG.Core/Chart/Events/ChartEventExtensions.cs
```diff
@@ -41,6 +41,171 @@ public static List<TNote> DuplicateNotes<TNote>(this List<TNote> notes)
...
+
+                notes.RemoveAt(index);
+
+                // Fix Previous/Next references
+                if (index == 0)
+                {
+                    // We were the first, so new 0's previous becomes null and everything else is fine
+                    notes[0].PreviousNote = null;
+                    return;
+                }
```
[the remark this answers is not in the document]
wyrdough: I added special handling above this for the case where this could fail. I believe that should address the concern.

YARG.Core/Chart/SongChart.AutoGeneration.cs
wyrdough: Reworked as suggested

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
wyrdough: Open note/chord handling reworked.

wyrdough:
Removed draft status since there is now an icon in the associated YARG PR.

YARG.Core/Chart/Events/ChartEventExtensions.cs
```diff
@@ -41,6 +42,179 @@ public static List<TNote> DuplicateNotes<TNote>(this List<TNote> notes)
...
+
+                if (index + 1 < notes.Count)
+                {
+                    newParent.NextNote = notes[index + 1];
+                    notes[index + 1].PreviousNote = newParent;
+                }
+
+                notes[index] = newParent;
+            }
```
sonicfind, line 181: Instead of constantly `note[index]`ing all over the place, cache the result of `notes[index]` at the top of the function to a local variable.

sonicfind: Then, when you have to do the `previousNote` & `NextNote` manipulation, use the local's own `previousNote` & `NextNote` values to do the work. No reason to check against the index when those values already tell you about the connections in both directions.

It's the standard pattern for when you have to remove a node from the middle of a linked list (which this is).

sonicfind: In fact, you can do away with all the extra `index > 0` and `index + 1 < count` checks entirely by just using the `previousNote` & `NextNote` variables.

YARG.Core/Chart/Events/ChartEventExtensions.cs
```diff
@@ -41,6 +42,179 @@ public static List<TNote> DuplicateNotes<TNote>(this List<TNote> notes)
...
+                return;
+            }
+
+            // If there's only one note with no children, we can just remove it and return, as there is
+            // nowhere left to move the flags and nothing to stitch back together
+            if (notes.Count == 1 && notes[index].ChildNotes.Count == 0)
+            {
+                notes.RemoveAt(index);
+                return;
+            }
```
sonicfind, line 61: Also, don't do this. This will always be the substantially uncommon path, so you should always aim to check for it as little as possible. Just let the logic underneath handle that case naturally.

sonicfind (changes requested):

YARG.Core/Chart/Events/ChartEventExtensions.cs
wyrdough: Seems more readable to me than protecting the notes[0] and notes[^1] later on....

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -55,58 +57,72 @@ public static void CompressGuitarRange(this InstrumentDifficulty<GuitarNote> dif
...
             int shiftAmount = rangeFrom - rangeTo;
-            GuitarNote note;
 
             for (var i = 0; i < difficulty.Notes.Count; i++)
             {
-                note = difficulty.Notes[i];
+                var note = difficulty.Notes[i];
+                bool isOpen = (GuitarEngine.OPEN_MASK & note.NoteMask) == GuitarEngine.OPEN_MASK;;
+                bool isOpenChord = (GuitarEngine.OPEN_MASK & note.NoteMask) > GuitarEngine.OPEN_MASK;
+                int openBit = GuitarEngine.OPEN_MASK & note.NoteMask;
```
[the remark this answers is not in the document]
wyrdough: Yeah, good thing it's actually unused (for the moment, anyway). Not sure how I missed that (or what made me think it would work in the first place)

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +36,215 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+                else
+                {
+                    note.NoteMask <<= -shiftAmount;
+                    note.DisjointMask <<= -shiftAmount;
+                }
+
+                // Fix up the open bit by clearing the open bit and then ORing with openBit, in case it was munged
+                // by a left shift
+                note.NoteMask = (note.NoteMask & ~GuitarEngine.OPEN_MASK) | openBit;
+                note.DisjointMask = (note.DisjointMask & ~GuitarEngine.OPEN_MASK) | openBit;
```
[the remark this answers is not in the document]
wyrdough: Why use a branch when it's not really necessary, tho? Seems readable enough to me, though maybe that will change a month from now, lol...

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +36,215 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+                        // We have found a sustain on the same fret as one of our notes, so we have to cut it off
+                        // at note.Tick
+                        if (prevNote.Tick + prevNote.TickLength >= parentOrChild.Tick)
+                        {
+                            prevNote.TickLength = parentOrChild.Tick - prevNote.Tick;
+                        }
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
wyrdough: I think I understand what you're saying would be a better way of going about it, but I don't quite understand the why behind it, aside from performance. (Which doesn't really matter much so long as it isn't absurdly slow given that we're not talking about charts with 20,000 notes here, just low hundreds in practice since it's only a thing on easy and medium charts and only runs at most once per player per song before the song actually starts)

Correct me if I'm wrong, but even an extended sustain can't overlap an intervening note in the same lane, so as long as we find the preceding note for each lane we should find the source of any sustain for that lane.

Or is the issue that you think that sustains shouldn't be cut and we should instead just drop the shifted note? Or something else entirely?

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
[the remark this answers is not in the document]
TheNathannator: Avoiding branches in this scenario is overall pennies in terms of performance optimization, I'd prefer readability over bitwise logic that in practice just obscures the actual intent somewhat (especially for something that runs at load time, where optimization is not a top priority).

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -34,6 +36,215 @@ public static void ConvertFromTypeToType(this InstrumentDifficulty<GuitarNote> d
...
+                // Check for validity of (possible) parent and remove if off track, reparenting children as necessary
+                var count = note.ChildNotes.Count + 1;
+                bool outOfRange = false;
+
+                foreach (var chordNote in note.AllNotes)
+                {
+                    if (chordNote.Fret is < (int) FiveFretGuitarFret.Green or > (int) FiveFretGuitarFret.Orange)
+                    {
+                        outOfRange = true;
+                    }
```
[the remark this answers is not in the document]
wyrdough: There's a fix in the latest push, though I'm sure that there is a better way to do it...

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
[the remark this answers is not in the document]
wyrdough: Changed as suggested

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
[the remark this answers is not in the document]
wyrdough: Reworked this to have separate hasOpen, isOpen, and isOpenChord bools. (Probably don't actually need isOpenChord, but I left it in case it turns out to be necessary and will clean it up once this takes its final form if it remains unused)

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -55,58 +57,72 @@ public static void CompressGuitarRange(this InstrumentDifficulty<GuitarNote> dif
...
 
-                if (rangeFrom == rangeTo)
-                {
-                    continue;
-                }
-
-                // Leave open notes as they are
-                if (note.Fret == (int) FiveFretGuitarFret.Open)
+                // No shifting required if from equals to or if the only note here is an open
+                if (rangeFrom == rangeTo || isOpen)
```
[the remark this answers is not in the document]
wyrdough: isOpen now correctly identifies whether the note is _only_ an open, so this should be fine now

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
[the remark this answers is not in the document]
wyrdough: So I think that the latest commit should be more along the lines of your suggested approach.

Basically what it does is creates an array of uints (that's actually bigger than the number of lanes for simplicity's sake, so that it can be indexed by the .Fret value) twiddles stuff, discards any notes that end up out of range or fail the note.tick < previous.tickend check, and then stores the end tick of whatever notes remain in an array before moving on to the next group of notes.

Works identically to previous for my test chart that has out of range notes, but I don't presently have one with an extended sustain that would result in a dropped note.

PR #251, opened 2025-06-02
Profile settings for drum highway customization

Frickitickitavi:
Backend changes for [YARG PR 1041](https://github.com/YARC-Official/YARG/pull/1041)


TheNathannator:
Had to be retargeted to engine-fixes last second since we realized this affects replay serialization.

PR #257, opened 2025-07-20
Make whammy star power gain frame independent

theli-ua:
This addresses 2 issues the current whammy tick gain calculations has:

1. CalculateStarPowerGain is dependent on how number of ticks are broken down 
    When working with the expression (quarterTick - lastQuarterTick) * (32.0 / 30.0) followed by Math.Round() to convert to a uint, there could be rounding inconsistencies depending on how the ticks are broken down.
    For example:
    If you have a difference of 30 ticks, you get 30 * (32.0/30.0) = 32.0, which rounds to 32
    If you break this into two segments of 15 ticks each, you get:
    15 * (32.0/30.0) = 16.0, which rounds to 16
    15 * (32.0/30.0) = 16.0, which rounds to 16
    Total: 32 (consistent)
    However, if you break it into segments of 10 ticks each:
    10 * (32.0/30.0) = 10.6666..., which rounds to 11
    10 * (32.0/30.0) = 10.6666..., which rounds to 11
    10 * (32.0/30.0) = 10.6666..., which rounds to 11
    Total: 33 (inconsistent with the single calculation)
    
2. When whammy is just started the first tick amount is calculated using a previous tick that was calculated at the previous frame's update. However that could mean that if more than 1 tick passed since then and whammy was activated somewhere in between it will/may overcount the amount of ticks.
     
number one is solved by carrying a remainder between frames. 
number two is solved using an explicit whammy start tick storage

With these changes I am no longer able to reproduce any SP gain differences with various frame timings

### DIFF
--- a/YARG.Core/Engine/BaseEngine.Generic.cs
+++ b/YARG.Core/Engine/BaseEngine.Generic.cs
@@ -40,6 +40,8 @@ namespace YARG.Core.Engine
 
         protected SustainList<TNoteType> ActiveSustains = new(10);
 
+        protected double WhammyTicksRemainder = 0.0;
+
         protected          int[]  StarScoreThresholds { get; }
         protected readonly double TicksPerSustainPoint;
         protected readonly uint   SustainBurstThreshold;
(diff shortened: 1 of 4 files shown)


YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -638,6 +651,11 @@ protected override void UpdateStarPower()
                 BaseStats.TotalStarPowerBarsFilled = (double) BaseStats.TotalStarPowerTicks / TicksPerFullSpBar;
                 YargLogger.LogFormatTrace("Gained {0} whammy ticks this update (Total: {1}), {2} sustains active. SP right now: {3}", whammyTicks, EngineStats.StarPowerWhammyTicks, ActiveSustains.Count, BaseStats.StarPowerTickAmount);
             }
+            else if (!StarPowerWhammyTimer.IsActive)
+            {
+                // TODO: if there is >= 0.5 there should it have been added somewhere?
```
theli-ua, line 656: This TODO is actually not valid, will remove. It's a leftover from previous iteration that used `floor` instead of `round`

PR #268, opened 2025-08-10
Save "AverageMultiplier" and "NumPauses" to Replay Stats

raphaelgoulart:
What it says on the tin. Also, it is done in a way that doesn't affect the replay file format.

### DIFF
--- a/YARG.Core/Engine/BaseStats.cs
+++ b/YARG.Core/Engine/BaseStats.cs
@@ -167,6 +167,16 @@ namespace YARG.Core.Engine
         /// </summary>
         public float Stars;
 
+        /// <summary>
+        /// The average multiplier of the player throughout the song.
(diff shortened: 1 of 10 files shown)


raphaelgoulart:
Fixed

TheNathannator:
Yeah, that's a bit of a tricky one. I'm not certain whether it should be serialized in BaseStats or if it should just be copied over from ReplayInfo upon deserialization.

I think I'll leave this PR to others for now, my time is pretty occupied lol

raphaelgoulart:
OK! Thanks for your work 🫡 I'll ping @wyrdough for this.

Also, since 0.13 _has_ been released since then, I think messing with the replay format for such a small change is undesirable, so I think my current solution is the ideal one given the circumstances.

PR #289, opened 2025-09-17
Configurable Rock Meter Presets / Improve Rock Meter for Vocalists

thomeval:
For details, see the associated PR on the YARG repo.

### DIFF
--- a/YARG.Core/Engine/EngineManager.FailMeter.cs
+++ b/YARG.Core/Engine/EngineManager.FailMeter.cs
@@ -177,11 +177,10 @@ namespace YARG.Core.Engine
                 {
                     OnHappinessOverThreshold?.Invoke();
                 }
-
-                // Send over threshold event when happiness goes from below threshold to above
-                if (Happiness >= HAPPINESS_CROWD_THRESHOLD && _previousHappiness < HAPPINESS_CROWD_THRESHOLD)
+                // Send under threshold event when happiness goes from above threshold to below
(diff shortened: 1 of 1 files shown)


YARG.Core/Engine/EngineManager.FailMeter.cs
```diff
@@ -73,70 +105,93 @@ public partial class EngineContainer
...
+                if (Happiness >= HAPPINESS_CROWD_THRESHOLD && _previousHappiness < HAPPINESS_CROWD_THRESHOLD)
+                {
+                    OnHappinessOverThreshold?.Invoke();
+                }
+
+                // Send over threshold event when happiness goes from below threshold to above
+                if (Happiness >= HAPPINESS_CROWD_THRESHOLD && _previousHappiness < HAPPINESS_CROWD_THRESHOLD)
+                {
+                    OnHappinessOverThreshold?.Invoke();
+                }
```
wyrdough, line 185: I think you intended for one of these to be over threshold and the other under, probably the following:

```suggestion
                if (Happiness >= HAPPINESS_CROWD_THRESHOLD && _previousHappiness < HAPPINESS_CROWD_THRESHOLD)
                {
                    OnHappinessOverThreshold?.Invoke();
                }
                // Send under threshold event when happiness goes from above threshold to below
                else if (Happiness < HAPPINESS_CROWD_THRESHOLD && _previousHappiness >= HAPPINESS_CROWD_THRESHOLD)
                {
                    OnHappinessUnderThreshold?.Invoke();
                }
```

wyrdough (changes requested):

PR #295, opened 2025-09-26
Support for normalization

polson:
Corresponds to YARG PR here: https://github.com/YARC-Official/YARG/pull/1146

### DIFF
--- a/YARG.Core/Audio/StemMixer.cs
+++ b/YARG.Core/Audio/StemMixer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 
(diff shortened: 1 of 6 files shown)


YARG.Core/Audio/StemMixer.cs
```diff
@@ -200,18 +219,6 @@ public int GetSampleData(float[] buffer)
             }
         }
 
-        public int GetLevel(float[] level)
```
wyrdough, line 203: I know it looks like we aren't using this, but it will be needed in the nearish future. (In case it makes a difference, the need is to be able to get the *output* level of the overall mix so that we can end songs with excessive trailing silence early)

PR #296, opened 2025-10-03
Band Score Multipliers

thomeval:
This PR adds support for Band Multipliers to Yarg.Core. For details, see the associated YARG repo PR.

### DIFF
--- a/YARG.Core/Engine/BaseStats.cs
+++ b/YARG.Core/Engine/BaseStats.cs
@@ -29,7 +29,7 @@ namespace YARG.Core.Engine
         /// <remarks>
         /// Calculated from <see cref="CommittedScore"/>, <see cref="PendingScore"/>, and <see cref="SoloBonuses"/>.
         /// </remarks>
-        public int TotalScore => CommittedScore + PendingScore + SoloBonuses;
+        public int TotalScore => CommittedScore + PendingScore + SoloBonuses + BandBonusScore;
 
         /// <summary>
         /// Total score earned from hitting notes.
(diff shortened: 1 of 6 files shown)


YARG.Core/Engine/BaseStats.cs
```diff
@@ -29,7 +29,7 @@ public abstract class BaseStats
         /// <remarks>
         /// Calculated from <see cref="CommittedScore"/>, <see cref="PendingScore"/>, and <see cref="SoloBonuses"/>.
         /// </remarks>
-        public int TotalScore => CommittedScore + PendingScore + SoloBonuses;
+        public int TotalScore => CommittedScore + PendingScore + SoloBonuses + BandBonusScore;
```
wyrdough, line 32: Please update the corresponding docstring to avoid future confusion

YARG.Core/Engine/EngineManager.FailMeter.cs
```diff
@@ -191,7 +191,12 @@ private void OnOverstrum()
 
             private void OnKeysOverhit(int key) => OnOverstrum();
 
-            private void OnStarpowerStatus(bool isActive) => _engineManager._starpowerCount += isActive ? 1 : -1;
+            private void OnStarPowerStatus(bool active)
```
wyrdough, line 194: I think for organizational purposes, this should be moved into `EngineManager.cs` since it is no longer specific to the fail meter. Same goes for `UpdateStarpowerCount()`.

It also feels slightly wrong to have all the event subscription stuff in FailMeter, but I'll take care of that when I eventually get the BRE stuff pulled in. I don't think it's worth addressing at this point.

wyrdough (changes requested):
Still need to do some functional testing to make sure, but I don't see anything that looks like it shouldn't work, just a few minor nitpicks to try and make our lives easier in the future...

YARG.Core/Engine/BaseStats.cs
thomeval: Resolved

YARG.Core/Engine/EngineManager.FailMeter.cs
thomeval: Resolved

PR #306, opened 2025-11-16
Add new menu vocal samples

W-Livi:
adding barks for the anti-piracy popup, the title screen, the settings and profiles menus, the quick keybinding screen, and the music library


wyrdough (approved):

polson:
I would be okay with these if we make vocals samples off by default

PR #316, opened 2026-01-14
Don't loop song previews

Ape:
The preview can be easily repeated by choosing the song again.


polson:
Now that we have advanced settings, what do you think about adding a setting for this?  Also I'd like some more description on how this works since I haven't tested it.  Does it just play the whole song from the beginning in the music library?

Ape:
> Does it just play the whole song from the beginning in the music library?

It plays the section that's usually looping, but just exactly once.

PR #325, opened 2026-02-07
Additional Beatles metadata support

trainlos:
Adds support for parsing the year from `date_released`, and also infers `artist` and `vocal_gender` for Beatles songs since these are not included in the DTA.
The case for `context` has been commented out because all it seemed to do was cause issues with "All You Need Is Love." Other than this song, every DTA I checked only had an integer for `context`.


polson (changes requested):

polson (approved):

polson:
Rebase this branch with master and I'll get it merged

PR #333, opened 2026-02-24
Drum highway configuration menu backend support

Frickitickitavi:
[Corresponding YARG PR](https://github.com/YARC-Official/YARG/pull/1351)


YARG.Core/Game/YargProfile.Obsolete.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        [Obsolete("Superseded by highway reordering.")]
+        public bool? SwapSnareAndHiHat;
+
+        [Obsolete("Superseded by highway reordering.")]
+        public bool? SwapCrashAndRide;
+
+
+#pragma warning disable 612, 618 // Ignore obsolete warnings since this is the place where we grandfather them in
+        public void GrandfatherIn()
```
wyrdough: Maybe call this `UpgradeDrumCustomization` or something?

wyrdough (changes requested):
Not absolutely certain that's everything, but it's what I have for now

YARG.Core/Game/YargProfile.Obsolete.cs
Frickitickitavi: Well, my thinking is that this is the general "grandfather in any obsolete settings" function, which just so happens to only cover these drum settings so far. I'd like it to be clear to the uninitiated developer that `YargProfile.cs` handles the current settings, and `YargProfile.Obsolete.cs::GrandfatherIn` is the entry point for all the legacy-handler code.

If you'd prefer, I can rename this function `UpgradeDrumCustomization`, and then create a new `GrandfatherIn` function that currently just calls `UpgradeDrumCustomization()`.

PR #344, opened 2026-03-21
Implement exPKG (PS3 format exCON) scanning and loading

Supersonicboss1:
I'm not sure this is exactly the best way to do this, since it first checks for the Xbox-style ex-cons and then only if those files don't exist, checks for the PS3-style ex-cons, which might affect performance when scanning the PS3 files.

The main differences between the Xbox and PS3 styles is the file names (.png_xbox -> .png_ps3, .milo_xbox -> .milo_ps3, .mid -> .mid.edat), the pngs have every byte swapped with its neighbour after the header, and that the midi files can be encrypted as .edats.

Since we don't want to support decrypting edats for obvious reasons, this just throws an error if the midi is detected as encrypted (First three bytes are NPD -> `4E 50 44`). Also, same as other CONs, moggs need to be unencrypted. RB3DX does support both unencrypted moggs and edats, so libraries should be compatible if set up correctly.

Also fixed a bug where .milo files were not being loaded at all for ex-cons.

### DIFF
--- a/YARG.Core/IO/Images/YARGImage.cs
+++ b/YARG.Core/IO/Images/YARGImage.cs
@@ -52,6 +52,15 @@ namespace YARG.Core.IO
             using var data = FixedArray.LoadFile(path);
             return TransferDXT(data);
         }
+        public static YARGImage LoadPS3DXT(string path)
+        {
+            using var data = FixedArray.LoadFile(path);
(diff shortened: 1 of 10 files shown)


polson (approved):

wyrdough (changes requested):
I don't see anything wrong with the functionality, but I don't think that adding conditionals to the exCON loader to look for the PS3-formatted names is really the way to go here, as it'll just be confusing in the future.

PR #346, opened 2026-03-28
Rework star and band star calculations

Supersonicboss1:
## Overview
This PR reworks both star and band star calculations to follow [this proposal](https://discord.com/channels/1086048856678084609/1107075236219850813). For a player, the main differences that will be noticed are:
1. In a band, SP now has a much larger effect on the band stars than before, since band bonus score is now taken into account
2. Solo bonuses are now taken into account in single-player.
3. Band stars are now based on total score, rather than being an average of all player's stars. This means players on higher difficulties, players with more notes in their chart (e.g. drums), and players on pro instruments (they get more points per note) will be factored into the star calculations more than before.

And yes, you can still gold star the children's keys packs (at least from the ones I tested). It does make gold stars in a band very difficult if the chart does not have unison sections, but this isn't an issue for official YARG charts from my testing.

Demo (I adjusted the bots to activate SP at the same time for this demo):
[2026-03-28 14-41-18.webm](https://github.com/user-attachments/assets/2c887f2f-9917-40cf-843d-d18617781d43)

## Technical changes

`StarScore` has been removed. It's only difference was it was total score without solo bonuses, but since solo bonuses are being factored in now, there is no reason for it to exist, except potentially for average multiplier calculation, and since that's only one use, it can just directly access `CommittedScore` and `PendingScore`.

The engines now take in an additional parameter, that being `SoloBonusStarMultiplierThresholds`, since this is needed to calculate the star score cutoffs. This is also serialised in the replays.

`BaseScore` calculation has been changed to not use a leniency factor, so it now represents the score if the player were to FC, hold all sustains fully , and never use SP. This will probably need to be updated when when `big-yarg-ending` is merged to take those into account too. This also fixes a bug with BaseScore calculations on Vocals that caused it to not take into account whether or not a phrase was percussion.

The function to calculate `BaseScore` has been changed to also calculate `BaseNoteScore`, which is the same, however does not factor in multiplier. This is intended to be used to calculate average multiplier for the score screen rework.

Much of the star calculation code has been moved to core, with YARG just passing in multiplier thresholds to the engine.

Calculation of Band star cutoffs, and updating of the band stars separately from player stars has been added to `EngineManager`. Band star updates are called in YARG, which I'm not super sure about, but it's the same way it was done before, so it should be fine? The code to save scores with bots in YARG has also been updated to match.

Replay version has also been bumped to 12. I think goulart wanted something in replays to tell if a player was a replay or not, so once I get that PR finished, they should probably be merged together to avoid bumping the version more than once. When loading old replays, the star multiplier cutoffs are also adjusted to match the new way they are calculated, due to the changes to `BaseScore`


YARG.Core/Engine/Drums/DrumsEngine.cs
```diff
@@ -359,12 +359,11 @@ protected override void AddScore(DrumNote note)
...
         }
 
-        protected sealed override int CalculateBaseScore()
+        protected sealed override (int baseScore, int noteScore) CalculateChartScores()
         {
-            double score = 0;
+            double baseScore = 0;
+            double noteScore = 0;
             int combo = 0;
-            int multiplier;
```
wyrdough, line 379: I believe this was being declared outside the loop to avoid having to shift the stack pointer several hundred times for no real reason. (also applies to the other engines)

YARG.Core/Engine/Keys/FiveLaneKeys/FiveLaneKeysEngine.cs
```diff
@@ -209,23 +208,21 @@ protected sealed override int CalculateBaseScore()
...
-
+                baseScore += multiplier * scoreForNote;
+                noteScore += scoreForNote;
                 // Pro Keys combo increments per chord, not per note.
                 combo++;
             }
 
-            YargLogger.LogDebug($"[Pro Keys] Base score: {score}, Max Combo: {combo}");
-            return (int) Math.Round(score);
+            YargLogger.LogDebug($"[Pro Keys] Base score: {baseScore}, Max Combo: {combo}");
```
wyrdough: The log statement is incorrect (oops), this is 5LK, not pro.

Also, shouldn't we be adding sustain score to 5LK like we do for guitar?

YARG.Core/Engine/Keys/ProKeys/Engines/ProKeysEngine.cs
```diff
@@ -243,23 +242,23 @@ protected sealed override int CalculateBaseScore()
...
+                    noteScore += pointsForSustain;
                 }
 
                 // Pro Keys combo increments per chord, not per note.
                 combo++;
             }
 
-            YargLogger.LogDebug($"[Pro Keys] Base score: {score}, Max Combo: {combo}");
-            return (int) Math.Round(score);
+            YargLogger.LogDebug($"[Pro Keys] Base score: {baseScore}, Max Combo: {combo}");
```
wyrdough, line 261: Shouldn't we be adding sustain score?

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -107,17 +106,24 @@ protected BaseEngine(InstrumentDifficulty<TNoteType> chart, SyncTrack syncTrack,
...
-            BaseScore = CalculateBaseScore();
+            (BaseScore, BaseNoteScore) = CalculateChartScores();
+
+            Solos = GetSoloSections();
+            Codas = GetCodaSections();
+            EngineStats.TotalSoloBonusPoints = CalculateTotalSoloBonus();
+
+            StarScoreThresholds = PopulateStarScoreThresholds(engineParameters.StarMultiplierThresholds, engineParameters.SoloBonusStarMultiplierThresholds, BaseScore, EngineStats.TotalSoloBonusPoints);
+        }
+        public static int[] PopulateStarScoreThresholds(float[] multiplierThresholds, float[] soloBonusMultiplierThresholds, int baseScore, int soloScore)
```
wyrdough, line 120: Missing a space between the two methods

YARG.Core/Engine/BaseEngineParameters.cs
```diff
@@ -47,7 +50,25 @@ protected BaseEngineParameters(ref FixedArrayStream stream, int version)
             StarMultiplierThresholds = new float[count];
             for (int i = 0; i < StarMultiplierThresholds.Length; i++)
             {
-                StarMultiplierThresholds[i] = stream.Read<float>(Endianness.Little);
+                // The way BaseScore is calculated changed in version 12, so the thresholds are stored differently (multiplied by the max multiplier) for older versions
+                int factor = version >= 12 ? 1 : MaxMultiplier;
```
wyrdough: We ended up on version 12 already for other reasons, so this will need to be updated to 13

YARG.Core/Engine/BaseEngineParameters.cs
```diff
@@ -47,7 +50,25 @@ protected BaseEngineParameters(ref FixedArrayStream stream, int version)
...
+                StarMultiplierThresholds[i] = stream.Read<float>(Endianness.Little) / factor;
+            }
+
+            if (version >= 12)
```
wyrdough: 13 here, too

YARG.Core/Engine/BaseStats.cs
```diff
@@ -185,6 +176,11 @@ public abstract class BaseStats
         /// </summary>
         public int SoloBonuses;
 
+        /// <summary>
+        /// Total number of bonus points available from solos. Should not be modified.
+        /// </summary>
+        public int TotalSoloBonusPoints;
```
wyrdough: I think it would be more clear if this were named `MaxSoloBonusPoints` or similar since it is unrelated to the player's performance.

YARG.Core/Replays/ReplayIO.cs
```diff
@@ -29,7 +29,7 @@ public static class ReplayIO
         private static readonly EightCC REPLAY_MAGIC_HEADER_OLD = new('Y', 'A', 'R', 'G', 'P', 'L', 'A', 'Y');
         private static readonly EightCC REPLAY_MAGIC_HEADER = new('Y', 'A', 'R', 'E', 'P', 'L', 'A', 'Y');
 
-        private static readonly (int OLD_MIN, int METADATA_MIN, int DATA_MIN, int CURRENT) REPLAY_VERSIONS = (4, 6, 9, 11);
+        private static readonly (int OLD_MIN, int METADATA_MIN, int DATA_MIN, int CURRENT) REPLAY_VERSIONS = (4, 6, 9, 12);
```
wyrdough: Needs to be updated to 13 since we've gone to 12 since the PR was authored...

wyrdough (changes requested):

YARG.Core/Engine/Drums/DrumsEngine.cs
Supersonicboss1: To be honest, I just saw the green squiggles in rider and followed its advice lol, but yeah, that was probably done for a reason.

YARG.Core/Engine/Keys/ProKeys/Engines/ProKeysEngine.cs
Supersonicboss1: We are?

wyrdough: I don't know how I missed that

wyrdough:
If you can fix up the unit test in `VocalsEngineTests.cs` I can merge this

PR #360, opened 2026-04-16
Revert drum BREs to single-zone, move visual BRE logic out of Core

Frickitickitavi:
[Corresponding YARG PR](https://github.com/YARC-Official/YARG/pull/1438)

### DIFF
--- a/YARG.Core/Engine/CodaSection.cs
+++ b/YARG.Core/Engine/CodaSection.cs
@@ -21,9 +21,6 @@ namespace YARG.Core.Engine
         // Last time bonus was collected for given lane
         private double[] LastCollectedTime { get; set; }
 
-        // We need this because collected and hit are different for non-guitar instruments
-        private double[] LastHitTime { get; set; }
-
         // Maximum bonus for one fret press or drum hit
         public int MaxLaneScore { get; private set; }
 
@@ -52,16 +49,15 @@ namespace YARG.Core.Engine
             return (float)Math.Min(visualTime - mostRecentTime, BONUS_RECHARGE_TIME);
         }
 
-        public CodaSection(int lanes, double startTime, double endTime, bool fretMode = true)
+        public CodaSection(int scoringZones, double startTime, double endTime)
         {
-            ScoringZones = lanes;
-            LastCollectedTime = new double[lanes];
-            LastHitTime = new double[lanes];
-            MaxLaneScore = fretMode ? MAX_FRET_SCORE : MAX_DRUM_SCORE;
+            ScoringZones = scoringZones;
+            LastCollectedTime = new double[scoringZones];
+            _fretMode = scoringZones == 1;
+            MaxLaneScore = _fretMode ? MAX_FRET_SCORE : MAX_DRUM_SCORE;
             TotalCodaBonus = 0;
             StartTime = startTime;
             EndTime = endTime;
-            _fretMode = fretMode;
             // MissNote will change this if necessary
             Success = true;
         }
@@ -106,8 +102,6 @@ namespace YARG.Core.Engine
                 TotalCodaBonus += bonusScore;
             }
 
-            LastHitTime[scoringZoneIndex] = time;
-
             OnLaneHit?.Invoke(fret);
         }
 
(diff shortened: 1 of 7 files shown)


YARG.Core/Engine/CodaSection.cs
```diff
@@ -45,18 +42,22 @@ public class CodaSection
         private const int MAX_FRET_SCORE  = 150;
 
         // Time taken for bonus to recharge after collection
-        private const double BONUS_RECHARGE_TIME = 1.5;
+        public const double BONUS_RECHARGE_TIME = 1.5;
+
+        public static float GetNormalizedTimeSinceLastHit(double visualTime, double mostRecentTime)
```
wyrdough, line 47: The return value of `GetNormalizedTimeSinceLastHit` was normalized to 0.0 to 1.0 because it gets passed to `LaneElement.SetEmissionColor`, which does the following:

```
var strength = 1 - Mathf.Pow(normalizedTime, 0.2f);
var newColor = _color * strength;

_innerMaterial.SetColor(EmissionColor, newColor);
```

I guess it doesn't *really* matter since `Color` components are clamped to a minimum of 0.0f anyway, but with the current calculation we'll end up having the lanes fully dark after only a second rather than the expected `BONUS_RECHARGE_TIME` seconds.

In any event, if the value isn't being normalized any more, the name should be changed to reflect that.

YARG.Core/Engine/Keys/FiveLaneKeys/YargFiveLaneKeysEngine.cs
```diff
@@ -369,7 +369,7 @@ private void HandleCodaFretChange(double time)
                 int button = 1 << i;
                 if ((pressed & button) != 0)
                 {
-                    coda.HitLane(time, i);
+                    coda.HitLane(time, i == 5 ? 6 : i); // Shift open back up, since HitLane is expecting the FiveLaneKeysAction index
```
wyrdough: Couldn't you just get rid of the bit math that shifts open from 6 to 5 in the first place?

wyrdough (changes requested):
I haven't reviewed the corresponding YARG PR yet, so that could conceivably implicate something here, but this will have to do until then...

YARG.Core/Engine/CodaSection.cs
Frickitickitavi: Oh that was unintentional - I meant to copy the normalization too. Fixed, name unchanged

YARG.Core/Engine/Keys/FiveLaneKeys/YargFiveLaneKeysEngine.cs
Frickitickitavi: Yes, if we extend the `for` loop up to `i < 7` and add `if (i == 5) continue` to the top of the loop. Not sure if that's any prettier, and the bit shift was here when I got here so I figured this was the less-invasive change

wyrdough: Unless something on the YARG side would get confused (I still haven't looked at it enough to know), you could force the wildcard bit to zero with something like the following and avoid needing a conditional inside the loop at all.

```
var wildcardBit = 1 << 5;
var pressed = (pressed & ~wildcardBit);
```

Frickitickitavi: Sure, that sounds good. While I'm in here, I think I _will_ move Wildcard to 7 after all - not that it matters for this approach, but I just can't stop thinking about how 6F's wildcards are going to be 7 and how annoying it'll be to have them mismatched lol

**Edit:** Actually, I guess it _does_ make a difference for this approach, since we don't even need to do the `~wildcardBit` thing if we stop short of 7 anyway. Even better then

wyrdough: Yes, that's even better

Frickitickitavi: Alright, pushed. Wildcard is now above open across FiveFretGuitarFret, FiveLaneKeysAction, GuitarAction, and even SixFretGuitarFret (which didn't have wildcard at all yet) for good measure. I've tested everything I can think of that might have broken as a result and it all seems good.

This does mean we're needlessly allocating one extra entry in FiveLaneKeysEngine::KeyPressTimes, but I think life will go on

wyrdough (approved):

PR #361, opened 2026-04-16
Fix nullable reference and unused field warnings

theli-ua:
just to get rid of the noise

- CONEntryGroup.cs: Changed _data to nullable (FixedArray<byte>?) and updated Dispose() to use null-conditional operator (CS8618: field not initialized in constructor, may be null)
- CONUpgradeGroup.cs: Changed _data to nullable in both PackedCONUpgradeGroup and UnpackedCONUpgradeGroup, updated Dispose() and null assignments (CS8618, CS8625: nullable reference handling)
- PackedCONEntryGroup.cs: Changed out parameter to nullable type and initialization to null! (CS8625: explicit null handling for out params)
- SongEntry.Loading.cs: Made Image property nullable (YARGImage?) (CS8603: _image may be null)
- YARGMiloReader.cs: Removed unused BlockInfo.Type field (CS0649: field never assigned)
- YargProKeysEngine.cs: Removed unused _keyPressedThisUpdate field (CS0169/CS0414: field declared but never used)


wyrdough:
I'd rather you suppress the warning in YARGMiloReader, as future extension may require actually reading the `Type` field.

theli-ua:
> I'd rather you suppress the warning in YARGMiloReader, as future extension may require actually reading the `Type` field.

Well, maybe we could just read it already? And *then* future extension may actually use it?

Otherwise I feel like suppression there would be kinda wrong. Although I'll totally do it if you insist.

wyrdough:
> > I'd rather you suppress the warning in YARGMiloReader, as future extension may require actually reading the `Type` field.
> 
> Well, maybe we could just read it already? And _then_ future extension may actually use it?
> 
> Otherwise I feel like suppression there would be kinda wrong. Although I'll totally do it if you insist.

Technically it is being read, it just isn't being set. Add `Type = type` to the initializer on line 128 and it will be set.

It still won't be used for anything, but given that the milo format isn't really documented it may turn out that we need to know which block type we're dealing with at some point in the future.

theli-ua:
> > > I'd rather you suppress the warning in YARGMiloReader, as future extension may require actually reading the `Type` field.
> > 
> > 
> > Well, maybe we could just read it already? And _then_ future extension may actually use it?
> > Otherwise I feel like suppression there would be kinda wrong. Although I'll totally do it if you insist.
> 
> Technically it is being read, it just isn't being set. Add `Type = type` to the initializer on line 128 and it will be set.
> 
> It still won't be used for anything, but given that the milo format isn't really documented it may turn out that we need to know which block type we're dealing with at some point in the future.

Sounds good

PR #362, opened 2026-04-18
Target dotnet 10.0 with the unit test and benchmark projects

jstnlef:
I suspect this would make it easier for people to run the tests locally since they can just download the most recent version of dotnet.


wyrdough (approved):
Hope I've got 10.x installed lol

PR #364, opened 2026-04-18
Change the the Audio PreviewContext to be controlled by it's caller

jstnlef:
This change is needed for https://github.com/YARC-Official/YARG/pull/1446

### DIFF
--- a/YARG.Core.UnitTests/Audio/PreviewContextTests.cs
+++ b/YARG.Core.UnitTests/Audio/PreviewContextTests.cs
@@ -97,6 +97,40 @@ public class PreviewContextTests
         Assert.That(mixer.DisposeCount, Is.EqualTo(1));
     }
 
+    [Test]
+    public async Task WaitForCompletionAsync_AllowsFadeOutAfterCancellation()
+    {
+        using var cancellationTokenSource = new CancellationTokenSource();
+        var manager = new FakeAudioManager();
+        var mixer = new FakeStemMixer(manager, length: 0.5);
+        var entry = new TestPreviewSongEntry(() => mixer);
+
+        var context = await PreviewContext.Create(
+            entry,
+            volume: 1f,
+            speed: 1f,
+            delaySeconds: 0,
+            fadeDuration: 0.05,
+            cancellationTokenSource.Token);
+
+        Assert.That(context, Is.Not.Null);
+
+        var waitTask = context!.WaitForCompletionAsync();
+        cancellationTokenSource.Cancel();
+
+        await Task.Delay(10);
+
+        using (Assert.EnterMultipleScope())
+        {
+            Assert.That(mixer.FadeOutCallCount, Is.EqualTo(1));
+            Assert.That(waitTask.IsCompleted, Is.False);
+        }
+
+        await waitTask;
+
+        Assert.That(mixer.DisposeCount, Is.EqualTo(1));
+    }
+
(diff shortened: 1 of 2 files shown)


YARG.Core/Audio/PreviewContext.cs
```diff
@@ -92,69 +98,78 @@ public class PreviewContext : IDisposable
...
-                    while (watch.Elapsed.TotalSeconds < _fadeDruation)
+                    _mixer.FadeOut(_fadeDuration);
+                    while (watch.Elapsed.TotalSeconds < _fadeDuration)
                     {
                         if (_disposed)
                         {
                             return;
                         }
-                        await Task.Delay(1);
+                        await Task.Delay(1, _token);
```
wyrdough: This should not be cancellable, as we want the preview to fade out when moving between tracks in the music library.
```suggestion
                        await Task.Delay(1);
```

YARG.Core/Audio/PreviewContext.cs
```diff
@@ -92,69 +98,78 @@ public class PreviewContext : IDisposable
...
                     watch.Restart();
-                    while (watch.Elapsed.TotalSeconds < _previewLength - _fadeDruation && !_token.IsCancellationRequested)
+                    while (watch.Elapsed.TotalSeconds < _previewLength - _fadeDuration && !_token.IsCancellationRequested)
                     {
                         if (_disposed)
                         {
                             return;
                         }
-                        await Task.Delay(1);
+                        await Task.Delay(1, _token);
```
wyrdough: If you make this one non-cancellable as well, I don't think we'll ever get an `OperationCanceledException` that we have to handle.


```suggestion
                        await Task.Delay(1);
```

YARG.Core/Audio/PreviewContext.cs
```diff
@@ -165,6 +180,10 @@ private async void Loop()
                     }
                 }
             }
+            catch (OperationCanceledException) when (_token.IsCancellationRequested)
+            {
+                Dispose();
+            }
```
wyrdough: Unless you choose to keep the `Delay` in the fade in loop cancellable (or I'm missing a case where we could otherwise get a cancellation throw), get rid of this.


```suggestion
```

YARG.Core.UnitTests/Audio/PreviewContextTests.cs
```diff
@@ -0,0 +1,212 @@
...
+            speed: 1f,
+            delaySeconds: 0,
+            fadeDuration: 0,
+            cancellationTokenSource.Token);
+
+        Assert.That(context, Is.Not.Null);
+
+        var stopTask = context!.WaitForCompletionAsync();
+        // ReSharper disable once MethodSupportsCancellation
+        await Task.Delay(25);
```
wyrdough, line 89: I suspect this will need to be adjusted to take into account the fade out delay (plus a millisecond)

wyrdough (changes requested):
Just a minor issue causing some undesired behavior in the music library

YARG.Core/Audio/PreviewContext.cs
jstnlef: Fixed

YARG.Core/Audio/PreviewContext.cs
jstnlef: Fixed

wyrdough (approved):

PR #365, opened 2026-04-19
﻿Fix unison phrase logic when starts are the same but ends are far apart

GenericMadScientist:
An example is the first phrases on guitar and bass of Misirlou GH2. The starts are exactly the same while the ends are more than a beat apart, but they are currently unison phrases.


wyrdough (approved):
Good catch, thanks for the PR!

PR #366, opened 2026-04-21
Fix natural HOPO state when HOPO follows a chord in the same lane

Supersonicboss1:
This fixes incorrect parser behaviour, that seems to only affect .chart, although in theory also applies to .mid, where notes that are within the hopo threshold, but are in the same lane as any of the chord frets, and incorrectly treated as hopos, where they should be strums. FWIW, both ChartForge and Moonscraper also have this incorrect behaviour.

See [issue 1 here](https://gist.github.com/elicwhite/27ffd91bb601446679bcab2869802620) and [paragraph 1 here](https://thenathannator.github.io/GuitarGame_ChartFormats/Chart-File-Formats/chart-format/Tracks/5-Fret-Guitar/#note-mechanics) for proof this is unintended.

This is done by just not caring if the previous note was a chord. We just check if the fret masks have no overlap, and if that is the case, it becomes a hopo, otherwise it does not.

Test chart:
[Test Chart 1.zip](https://github.com/user-attachments/files/26937652/Test.Chart.1.zip)

### DIFF
--- a/YARG.Core/MoonscraperChartParser/Events/MoonNote.cs
+++ b/YARG.Core/MoonscraperChartParser/Events/MoonNote.cs
@@ -278,7 +278,7 @@ namespace MoonscraperChartEditor.Song
             // Checking state in this order is important
             return !isChord &&
                 previous != null &&
-                (previous.isChord || rawNote != previous.rawNote) &&
+                (mask & previous.mask) == 0 &&
                 tick - previous.tick <= hopoThreshold;
         }
 


jstnlef:
Is it possible to add a test for this? :) If not for this one, I'm wondering if I should make a C# translation of the tool they were using to include to our test suite

Supersonicboss1:
I'm actually not even sure if this is the correct behavior anymore lol. Going to mark this as draft until I know for sure

TheNathannator:
To my knowledge, for .chart the overall correct behavior is to *not* consider the previous note's lanes, and vice versa for .mid. One of a few small inconsistencies between the two formats.

This matches the behavior observed in CH v1.0, last I checked. Unsure if this handling has changed since then.

Supersonicboss1:
> To my knowledge, for .chart the overall correct behavior is to _not_ consider the previous note's lanes, and vice versa for .mid. One of a few small inconsistencies between the two formats.
> 
> This matches the behavior observed in CH v1.0, last I checked. Unsure if this handling has changed since then.

Yeah, you are correct I think, and this matches the current behavior in YARG too - closing this PR.

Also, @jstnlef, if you wanna write tests for this case, here is an identical .mid and .chart, which tests this behaviour. On .chart, both single notes should be HOPOs, and on .mid, both should be strums.

[testcharts.zip](https://github.com/user-attachments/files/26941540/testcharts.zip)

PR #367, opened 2026-04-21
Add ChartDump CLI: dump MoonSong + SongChart to JSON

elicwhite:
## Summary

Adds **ChartDump**, a small CLI that parses a `.chart`/`.mid` file with YARG.Core's MoonSong parser and emits the parsed state as JSON. It dumps both:

- the **raw MoonSong** representation — per-track `notes[]` with `{tick, rawNote, length, flags, flagNames}`, special phrases, text events, animations, venue, etc. (pre-resolve — force flags still distinct from final HOPO/TAP state);
- the **normalized SongChart** representation — post-resolve guitar/drum/vocal notes, vocals phrases, drum dynamics/flags.

Two modes:

```
# single chart, JSON to stdout
dotnet run --project ChartDump -- <path-to-chart-folder-or-file>

# batch: walk input tree, write yarg-dump.json per chart folder
dotnet run --project ChartDump -- --batch <input-dir> <output-dir> [limit]
```

Batch mode runs with `MaxDegreeOfParallelism=16` and only writes `yarg-dump.json`, so it composes with an existing output tree without touching other files.

## Motivation

I wrote this to diff YARG.Core's parser against another parser across a large chart corpus, to isolate exact behavioral differences — which cases turn into a natural HOPO vs. strum, which SP/solo phrases get dropped vs. extended, how overlapping force modifiers resolve, etc. It's also useful as a single-chart debugger when triaging a reported parsing issue: you get the full parsed state without having to launch the game.

I think the tool has enough reuse value for YARG.Core maintainers and contributors that it belongs alongside `TestConsole`, `ReplayCli`, `YARG.Core.Benchmarks`, and `YARG.Core.UnitTests`. Happy to drop it back out if you'd prefer it live elsewhere.

## Changes

- `ChartDump/{ChartDump.csproj, Program.cs, DumpAll.cs}` — new console project. `net8.0` (matches `YARG.Core.UnitTests` / `YARG.Core.Benchmarks`), `RollForward=Major` so it also runs on newer SDKs.
- `YARG.Core.sln` — register the new project.
- `YARG.Core/Assembly.cs` — add `[assembly: InternalsVisibleTo("ChartDump")]`. `MoonSong`, `MidReader`, `ChartReader`, `ParseSettings`, etc. are internal, and ChartDump needs them for the raw-MoonSong dump layer. Same mechanism that `YARG.Core.Benchmarks` and `YARG.Core.UnitTests` already use.
- `YARG.Core/Game/Presets/ColorProfile.FiveLaneDrums.cs` — remove a stale `using DG.Tweening;`. Nothing in the file references the namespace, and its presence blocks non-Unity builds. Happy to split this into its own PR if you'd rather.

## Output shape (abbreviated)

```jsonc
{
  "resolution": 480,
  "format": "mid",
  "tempos": [...],
  "timeSignatures": [...],
  "sections": [...],
  "globalEvents": [...],
  "venue": [...],            // only if present
  "tracks": [                // raw MoonSong per instrument × difficulty
    {
      "instrument": "Guitar", "difficulty": "expert", "gameMode": "Guitar",
      "notes":   [{ "tick": 86880, "rawNote": 3, "length": 0, "flags": 0, "flagNames": null }, ...],
      "phrases": [{ "tick": 0, "length": 3840, "type": "Starpower" }, ...],
      "textEvents": [...],
      "animations": [...]
    }
  ],
  "vocalsTracks": { ... },   // normalized vocals from SongChart
  "normalizedTracks": [ ... ] // post-resolve guitar/drum notes
}
```

Drum-track selection respects `song.ini` — `five_lane_drums` / `pro_drums` hints are read and forwarded to `ParseSettings.DrumsType`, and only the canonical drum track is emitted in `normalizedTracks` (YARG populates all three interpretations from the same raw source; without the ini hint the diff would just see pad-index churn between non-canonical interpretations). HOPO threshold overrides (`hopo_frequency`, `eighthnote_hopo`) are also read from `song.ini` and applied to `ParseSettings.HopoThreshold`.

## Test plan

- [x] `dotnet build YARG.Core.sln -c Release` — succeeds on net8.0 via net10 SDK.
- [x] Single-chart mode produces non-empty JSON on a representative `.mid` chart.
- [x] `--batch` mode over a large corpus writes `yarg-dump.json` per folder without overwriting other files and without side effects on the source tree.
- [ ] Would be nice for maintainers to run the build on their CI target if there is one — I couldn't find a YARG.Core CI workflow that builds the solution, so I haven't added a CI job here; happy to if you'd like one.

### DIFF
--- a/ChartDump/ChartDump.csproj
+++ b/ChartDump/ChartDump.csproj
@@ -1,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <LangVersion>12</LangVersion>
+    <Nullable>enable</Nullable>
+    <RollForward>Major</RollForward>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\YARG.Core\YARG.Core.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/ChartDump/DumpAll.cs
+++ b/ChartDump/DumpAll.cs
@@ -1,0 +1,121 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Threading;
+using System.Threading.Tasks;
+
+/// <summary>
+/// Batch mode: walks a chart directory, parses each chart with YARG.Core,
+/// and dumps JSON files to an output directory matching the folder structure.
+///
+/// Usage: ChartDump --batch <input-dir> <output-dir> [limit]
+/// </summary>
+static class DumpAll
+{
+    public static int Run(string inputDir, string outputDir, int limit)
+    {
+        if (!Directory.Exists(inputDir))
+        {
+            Console.Error.WriteLine($"Input directory not found: {inputDir}");
+            return 1;
+        }
+
+        Directory.CreateDirectory(outputDir);
+
+        var folders = FindChartFolders(inputDir);
+        if (limit > 0 && limit < folders.Count)
+            folders = folders.GetRange(0, limit);
+
+        int success = 0, failed = 0, skipped = 0;
+        var startTime = DateTime.UtcNow;
+
+        var jsonOptions = new JsonSerializerOptions
+        {
+            WriteIndented = false,
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+            Converters = { new JsonStringEnumConverter() }
+        };
+
+        Parallel.ForEach(folders, new ParallelOptions { MaxDegreeOfParallelism = 16 }, folder =>
+        {
+            var relPath = Path.GetRelativePath(inputDir, folder);
+            var outPath = Path.Combine(outputDir, relPath, "yarg-dump.json");
+
+            try
+            {
+                var chartFile = Path.Combine(folder, "notes.chart");
+                if (!File.Exists(chartFile))
+                    chartFile = Path.Combine(folder, "notes.mid");
+                if (!File.Exists(chartFile))
+                {
+                    Interlocked.Increment(ref skipped);
+                    return;
+                }
+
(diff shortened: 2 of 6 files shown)


elicwhite:
@(the-commenter) — you're right, and the cause is on my end: ChartDump was using `ParseSettings.Default_Midi` / `Default_Chart` verbatim, but YARG's runtime doesn't. `SongEntry.IniBase.LoadChart` builds its own `ParseSettings` with `ChordHopoCancellation = _chartFormat != ChartFormat.Chart` — i.e. **true for .mid**. That flag is what drives the `MoonSongLoader.Guitar.GetGuitarNoteType` post-pass at lines 92–107: when a single note follows a chord within the HOPO threshold and shares a fret with the chord, the resolved type flips Hopo→Strum. Without setting the flag, ChartDump's `normalizedTracks` types diverged from what the game actually plays.

I also hit a second bug while chasing this one: ChartDump was feeding `DrumsType.ProDrums` straight from a `pro_drums` song.ini hint into `settings.DrumsType`, but `MoonSongLoader.Drums` only accepts `FourLane` / `FiveLane` (YARG's own `ParseDrumsType` never returns `ProDrums`). The result was `SongChart.FromFile` throwing `InvalidOperationException: Unexpected drums type ProDrums! (Drums type should have been calculated by now)` — which I was silently catching, so `normalizedTracks` was just quietly empty on every pro-drums chart.

Both fixed in 86f3f9c on the branch. Verified on `.38 Special - Hold On Loosely` guitar/expert: the tick-281040 Green note (following a `Forced_Hopo` Green+Red chord at tick 280880) now emits `type=Strum` instead of `type=Hopo`, matching in-game behavior. Same result on the other charts I had been using as test cases. Also switched the caught exception to log to stderr instead of swallowing — the silent catch is what hid the drums bug.

Separately, this invalidates a chunk of writing I'd done that was pointing at YARG's `IsNaturalHopo` as a parser bug — it's not, it's just that the chord-subset-is-strum rule lives in a different place than I was reading. Apologies for the misdirection in the PR description; I'll update the corresponding write-up on my end.

PR #369, opened 2026-04-24
Fix for "'.' in CON file name will break background video support"

jstnlef:
Fixes the bug reported here -> https://discord.com/channels/1086048856678084609/1496999114817273866


wyrdough:
My assumption when I first saw this was that the problem was `Path.GetFileNameWithoutExtension` being stupid, but you're still using it so I guess not?

jstnlef:
> My assumption when I first saw this was that the problem was `Path.GetFileNameWithoutExtension` being stupid, but you're still using it so I guess not?

The problem is that a con file with the name B.Y.O.B will treat the trailing .B as the extension. So it would strip it before looking for the .mp4 file

PR #370, opened 2026-04-24
Add ability to reset state of EngineManager

Supersonicboss1:
This was required for my replay fixes PR. This just adds a method to EngineManager and UnisonEvent that resets any data which encodes the current state of gameplay, similar to the methods on the engines.


wyrdough:
I don't think this is a complete fix since it can't rewind to a specific time and have the correct stats, but it's better than what we have now which is pretty scuffed, so I'm going to go ahead and merge.

PR #373, opened 2026-04-27
Six Fret Guitar Support

theli-ua:
Adds six fret guitar engine. The approach taken is essentially reusing five fret engine with anchoring and hopo rules override. 
This means that engine preset is shared.
Some refactoring was also required in existing five fret code. 
Colors are separate in color profile though.
Chart parsing was already there.

Also added ability to play five fret charts as well as requested, however I personally think that is worse than nothing since it would map.into forbidden chords


wyrdough:
All seems reasonable here as well...

Supersonicboss1 (changes requested):
Everything seems fine, aside from my issue with the general architecture I mentioned on Discord. But again, I won't make you fix it; you already had to wait for far too many months for this to be approved lol.

I should also note for anyone else that the *code* looks fine, I do not own a 6F guitar, so my testing was limited to what I could test with a normal 5F guitar.

The actual issue I have is that 6F has no modifiers. All of the modifiers, to my knowledge, just use `InstrumentDifficulty<GuitarNote>`, so most should be immediately compatible, they just need to be marked as such in `YargProfile` and `Modifier`, I believe. (translation strings will also need to be added on the YARG side). Some won't be compatible, e.g. Opens to Greens, but feel free to leave that - I plan to rework the modifiers at some point soon anyway.

theli-ua:
>  but feel free to leave that - I plan to rework the modifiers at some point soon anyway.

So you do not insist it be a part of this PR? 

I honestly completely forgot modifiers since I've never actually used them

YARG.Core/Game/YargProfile.cs
```diff
@@ -380,42 +380,49 @@ public void RestoreSavedModifiers()
...
                     }
                     else if (IsModifierActive(Modifier.TapsToHopos))
                     {
-                        guitarTrack.ConvertFromTypeToType(GuitarNoteType.Tap, GuitarNoteType.Hopo);
-                    }
-                    else if (IsModifierActive(Modifier.RangeCompress))
-                    {
-                        guitarTrack.CompressGuitarRange();
+                        guitarTrack!.ConvertFromTypeToType(GuitarNoteType.Tap, GuitarNoteType.Hopo);
                     }
```
Supersonicboss1, line 420: I don't like the use of `goto` here, I think it makes the code confusing. Could you extract the guitar modifier application into a seperate function, something like this:
```csharp
switch (GameMode) {
    case GameMode.FiveFretGuitar:
    case GameMode.SixFretGuitar:
        ApplyGuitarModifiers()
    // rest of the switch...
}

private void ApplyGuitarModifiers() {
     if (GameMode is GameMode.FiveFretGuitar)
         // 5F exclusive modifier code and checks
    
    // rest of the shared modifiers
}
```

Supersonicboss1 (changes requested):
sorry, one minor nitpick and we should be good to go on the core side.

YARG.Core/Game/YargProfile.cs
theli-ua: I've instead pretty much reverted this file and instead made that guitar branch apply to both gutiar fret modes.
It does not matter if, say, opens to greens is not one of valid six fret modes as it'll be filtered out in Modifier.cs and there is no way for the user to enable it.

Supersonicboss1 (approved):
everything looks good - you may want to check out the null warnings on `TryGetFive/SixFretDifficulty` in SongChart, though I don't think it'll cause issues in practice.

theli-ua:
> everything looks good - you may want to check out the null warnings on `TryGetFive/SixFretDifficulty` in SongChart, though I don't think it'll cause issues in practice.


I'm not familiar with those nullable concept and overall not very familiar with C#, for this one I've just accepted what llm convinced me to do (see latest commit)

Supersonicboss1:
> > everything looks good - you may want to check out the null warnings on `TryGetFive/SixFretDifficulty` in SongChart, though I don't think it'll cause issues in practice.
> 
> 
> 
> 
> 
> I'm not familiar with those nullable concept and overall not very familiar with C#, for this one I've just accepted what llm convinced me to do (see latest commit)

Honestly, I didn't know that annotation existed, but that seems correct, thanks!

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
```diff
@@ -7,6 +8,486 @@ namespace YARG.Core.Chart
...
+                    Build(index + 1, value + 1);
+                }
+            }
+
+            Build(0, 1);
+            return results;
+        }
+
+        /// <summary>
+        /// Removes the Orange member of a 5-note chord so the remainder can be placed legally.
```
polson: Description seems wrong here?

YARG.Core/Game/Modifier.cs
polson: You already have this case so this is unreachable

YARG.Core/Engine/Guitar/Engines/YargSixFretGuitarEngine.cs
```diff
@@ -0,0 +1,302 @@
...
+        /// </summary>
+        private const int FRET_BUTTON_MASK = 0x3F;
+
+        public YargSixFretGuitarEngine(InstrumentDifficulty<GuitarNote> chart, SyncTrack syncTrack,
+            GuitarEngineParameters engineParameters, bool isBot)
+            : base(chart, syncTrack, engineParameters, isBot)
+        {
+        }
+
+        protected override int GetChordLowestFretMask(GuitarNote note)
```
polson, line 69: Can you double check if the robit found a logic issue here?  I think the issue is physical fret and bit index are not the same like they are for five fret:

> `note.NoteMask` has bits 2 (Black 3) and 3 (White 1) set.
> 
> The for loop iterates fret (bit position) from 0 to 5:
> 
> fret = 0 (Black 1, bit 0): Not in chord.
> fret = 1 (Black 2, bit 1): Not in chord.
> fret = 2 (Black 3, bit 2): Match found! The loop executes break.
> It returns 1 << 2 (Black 3, fret 3).
> 
> Because White 1 (fret 1) is at bit 3, the loop never reaches it. It returns fret 3 instead of fret 1.

polson:
I am wondering if we should bump `COLOR_PROFILE_VERSION ` to 3 since we made changes to the color profile stuff?

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
theli-ua: Oops, I was sure I removed that

YARG.Core/Game/Modifier.cs
theli-ua: Right, forgot to remove when I actually added valid modifiers.

YARG.Core/Chart/Tracks/InstrumentDifficultyExtensions.cs
theli-ua: There was a method here at some point that I removed, that left this orphaned docstring. 

Fixed

YARG.Core/Engine/Guitar/Engines/YargSixFretGuitarEngine.cs
theli-ua: Indeed, that was meant to be overriden to be position based, along with isAnchoringValid

I've added a test for anchoring chords that covers a missed test and fixed this

theli-ua:
> I am wondering if we should bump `COLOR_PROFILE_VERSION ` to 3 since we made changes to the color profile stuff?

I'm not sure. 

Change is fully additive - so it should load in the old version and old settings should also load in the new code, albeit with defaults for white/black, right?

polson (approved):

PR #374, opened 2026-04-27
Support for reducing flashing lights

polson:
Support for https://github.com/YARC-Official/YARG/pull/1402

### DIFF
--- a/YARG.Core/Chart/Events/ChartEvent.cs
+++ b/YARG.Core/Chart/Events/ChartEvent.cs
@@ -38,10 +38,11 @@ namespace YARG.Core.Chart
         }
 
         /// <summary>
-        /// Walks the list, comparing each event against the last kept event.
-        /// Skips events where shouldSkip returns true.
+        /// Filters events by enforcing a minimum interval between them,
+        /// and dropping duplicates detected by the given predicate.
         /// </summary>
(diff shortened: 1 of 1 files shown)


YARG.Core/Chart/Events/ChartEvent.cs
```diff
@@ -33,5 +36,34 @@ protected bool Equals(ChartEvent other)
...
+                return events;
+            }
+
+            var filtered = new List<T>(events.Count) { events[0] };
+            for (var i = 1; i < events.Count; i++)
+            {
+                var current = events[i];
+                var previous = filtered[^1];
+                var isWithinInterval = current.Time - previous.Time < interval;
+                var shouldSkipDuplicate = isDuplicate(current, previous);
```
wyrdough: This is going to nerf smooth transitions in some cases, but it isn't immediately obvious to me how to easily account for that, so I'm not too bothered...

For context, the way that blending works (I believe it's the same for both lighting and postproc) is that if we are on effect type A and then we encounter another effect type A, we initiate a blend to the next effect type (call it B) with a duration of the time between the second A and B. Sometimes, charts have A1 simultaneous with A2 to create the effect of switching from the previous effect to A and immediately starting a blend to B.

polson: Updated to not filter events with the exact same time to support fading

wyrdough (approved):
Didn't test, but I'm happy with this

PR #378, opened 2026-05-03
Adding a few more engine tests

jstnlef:
This PR actually fixes a couple of small bugs as well around some boundaries checking and what I believe to be missing properties on the BaseStats copy.

### DIFF
--- a/YARG.Core/Engine/EngineManager.Band.cs
+++ b/YARG.Core/Engine/EngineManager.Band.cs
@@ -1,18 +1,20 @@
 ﻿using System;
 using System.Collections.Generic;
+using YARG.Core.Logging;
 
 namespace YARG.Core.Engine
 {
     public partial class EngineManager
     {
+        private const int NUMBER_OF_STAR_SCORE_THRESHOLDS = 6;
         public int   Score { get; set; }
         public int   Combo { get; set; }
         public float Stars { get; private set; }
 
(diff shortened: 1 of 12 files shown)


YARG.Core/Engine/EngineManager.Band.cs
```diff
@@ -87,9 +89,19 @@ private void UpdateBandMultiplier()
 
         public static int[] GetStarScoreCutoffs(List<int[]> starScoreCutoffsList)
         {
+#if DEBUG
```
wyrdough: This should probably be `#if UNITY_EDITOR || YARG_TEST_BUILD || YARG_NIGHTLY_BUILD`. (Maybe not nightly if I'm misremembering and this gets called frequently during gameplay)

wyrdough (changes requested):
Sorry for the delay in getting to this..

YARG.Core/Engine/EngineManager.Band.cs
jstnlef: I'm happy to make the change. It is only called in 2 places. Once on song loading and once when the game goes to record the scores after the song

jstnlef: All set 👍

wyrdough (approved):

PR #380, opened 2026-05-13
Add average multiplier to BaseStats

Supersonicboss1:
This was required for my score screen replay PR. It was kind of just... a pain to access AverageMultiplier when it was somewhere else, or calculated in the score screen, so I moved it here. Not serialised, because it's in ReplayStats anyway (where it is actually saved now, hooray!) and there's no reason to bump the replay version for something you can derive anyway.

This is calculated in AddScore. I don't know if there's anything we could do with average multiplier during gameplay, but I figured it could be a useful stat at some point?

### DIFF
--- a/YARG.Core/Engine/BaseEngine.Generic.cs
+++ b/YARG.Core/Engine/BaseEngine.Generic.cs
@@ -753,6 +753,8 @@ namespace YARG.Core.Engine
             // scoreMultiplier includes combo+star power score
             EngineStats.CommittedScore += scoreMultiplier;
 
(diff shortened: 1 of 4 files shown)


wyrdough:
If it's needed for replays, shouldn't we be calculating and setting when deserializing?

Supersonicboss1:
> If it's needed for replays, shouldn't we be calculating and setting when deserializing?

Yeah, we probably should. Though this does make me think that it might be better to make this a part of BaseStats in replays, and remove it from ReplayStats. It's not like it was even being recorded before now. It would require a replay version bump, but we do need to do that anyway for the autohit lane PR (and for removing NumPauses). Thoughts?

PR #382, opened 2026-05-15
Stop counting percussion phrases in TotalNotes

Supersonicboss1:
Fixes the issue where it was impossible to FC a vocals chart with percussion, since the percussion phrase counted as a phrase, but didn't increment combo.

I could have added a special case to the BaseEngine constructor but it felt weird to do that since I'd need to check if it was of type `VocalsNote` first. I don't think there are any adverse effects to not calling `OnNoteHit`/`IncrementNotesHit` on them, since percussion phrases don't really mean anything as far as I can tell, so this is essentially just not counting them as notes.


wyrdough (commented):
Does this not prevent the hit animation from playing?

Supersonicboss1:
> Does this not prevent the hit animation from playing?

No, that still works. I'm only not invoking `OnPercussionHit` on the note that is a percussion **phrase**, not an actual percussion hit, and everything seems to still work as expected.

PR #383, opened 2026-05-16
Autohit-centric lane logic

Frickitickitavi:
[Corresponding YARG PR](https://github.com/YARC-Official/YARG/pull/1491)

This PR changes the fundamentals of lanes work:
* Within a lane, overhit protection (for the inputs that the lane expects) is never revoked under any circumstances
* `LaneExpireTime` and `TremoloWindow` no longer exist per se, but are closely replaced by `LaneAutohitExpireTime` and `LaneAutohitWindow`
* `LaneAutohitExpireTime` is set to `CurrentTime + LaneAutohitWindow` whenever a lane receives an input, regardless of whether that input hit a note - as long as you're spamming inputs to the lane, you're keeping your autohit up
* When a laned note would pass out of the back of the hit window, we check if it's within the current autohit time and autohit it if so
* There is also a new `LaneProximityProtectionWindow` parameter (bit of a mouthful, feel free to suggest a rename) that drives autohit protection immediately before and after a lane (the latter was previously driven by the expire time, but is now flat from the end of the lane regardless of inputs)
  * On Drums and Keys, these protection windows only apply to inputs that would have satisfied the lane (e.g. you can't be forgiven for overhitting a Gcym right after a snare roll)
  * On Guitar, these protection windows apply to any overstrum at all, to provide forgiveness to players who might be transitioning between two tremolos and briefly fretting some intermediate state. They also protect against ghosting, but only for trills, and only for inputs that match the trill that's providing the forgiveness

The most important change here is that we no longer revoke overhit protection during lanes, and instead only revoke autohitting (a.k.a. underhit protection). This matches RB and CH's handling of lanes (despite what the RBN docs will tell you) and allows for lanes to be charted more sparsely than they would otherwise need to be to avoid unfair overhits.

Aside from that, these changes also produce a happy middle-ground between RB and CH's systems:
* RB's system (hardcoded input speed requirement for literally any lane) enables lanes to preserve playability for any charted NPS (which many RB charts take advantage of), but makes the contents of the lane completely irrelevant
* CH's system (forgive every-other note in the lane) keeps the contents of the lane relevant to gameplay, but hits limitations with particularly fast lanes (found in many RB charts) where even half of the charted input speed is oppressive
* This system keeps the charted speed relevant to gameplay like CH, but also introduces a limit on the required speed to keep everything playable like RB. The `LaneAutohitWindow` engine parameter allows players to configure that speed limit if they find it inadequate or overly generous

Some notes for the reviewer:
* I'm unsure if I correctly updated the replay and engine version numbers
* I've tested these changes in gameplay, but not super robustly and will continue to verify that things work as intended
* I'm not married to any of the parameter names or tooltips, if you want to editorialize
* I'm also not married to the default values for the parameters; happy to retune
* This also sneaks in a few `BaseEngine` replay consistency fixes suggested by @jstnlef; see his comments for details

### DIFF
--- a/YARG.Core/Engine/BaseEngine.Generic.cs
+++ b/YARG.Core/Engine/BaseEngine.Generic.cs
@@ -55,8 +55,6 @@ namespace YARG.Core.Engine
         public override BaseEngineParameters BaseParameters => EngineParameters;
         public override BaseStats            BaseStats      => EngineStats;
 
-        protected const double LANE_PROXIMITY_LENIENCY = 0.080;
-
         protected BaseEngine(InstrumentDifficulty<TNoteType> chart, SyncTrack syncTrack,
             TEngineParams engineParameters, bool isChordSeparate, bool isBot)
             : base(syncTrack, isChordSeparate, isBot)
@@ -679,8 +677,8 @@ namespace YARG.Core.Engine
 
         protected void UpdateLaneAutohitExpireTime()
         {
-            LaneAutohitExpireTime = CurrentTime + EngineParameters.HitWindow.TremoloWindow;
-            YargLogger.LogFormatDebug("LaneExpireTime extended to {0}. TremoloWindow {1}. Increment {2}.", LaneAutohitExpireTime, EngineParameters.HitWindow.TremoloWindow, LaneAutohitExpireTime - CurrentTime);
+            LaneAutohitExpireTime = CurrentTime + EngineParameters.HitWindow.LaneAutohitWindow;
+            YargLogger.LogFormatDebug("LaneExpireTime extended to {0}. LaneAutohitWindow {1}. Increment {2}.", LaneAutohitExpireTime, EngineParameters.HitWindow.LaneAutohitWindow, LaneAutohitExpireTime - CurrentTime);
         }
 
         protected bool SkipPreviousNotes(TNoteType current)
--- a/YARG.Core/Engine/BaseEngine.Generic.cs
+++ b/YARG.Core/Engine/BaseEngine.Generic.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using YARG.Core.Chart;
@@ -55,7 +55,7 @@ namespace YARG.Core.Engine
         public override BaseEngineParameters BaseParameters => EngineParameters;
         public override BaseStats            BaseStats      => EngineStats;
 
-        protected const double LANE_END_LENIENCY = 0.080;
+        protected const double LANE_PROXIMITY_LENIENCY = 0.080;
 
         protected BaseEngine(InstrumentDifficulty<TNoteType> chart, SyncTrack syncTrack,
             TEngineParams engineParameters, bool isChordSeparate, bool isBot)
@@ -339,9 +339,9 @@ namespace YARG.Core.Engine
 
             if (IsLaneActive)
             {
-                if (IsTimeBetween(LaneExpireTime, previousTime, nextTime))
+                if (IsTimeBetween(LaneAutohitExpireTime, previousTime, nextTime))
                 {
-                    QueueUpdateTime(LaneExpireTime, "Potential Lane Expiration Time");
+                    QueueUpdateTime(LaneAutohitExpireTime, "Potential Lane Expiration Time");
                 }
             }
         }
@@ -359,6 +359,8 @@ namespace YARG.Core.Engine
 
         protected override void UpdateTimeVariables(double time)
         {
+            YargLogger.LogTrace($"REQUIRED LANE NOTE: {RequiredLaneNote}");
+
             if (time < CurrentTime)
             {
                 YargLogger.FailFormat("Time cannot go backwards! Current time: {0}, new time: {1}", CurrentTime,
@@ -415,17 +417,6 @@ namespace YARG.Core.Engine
 
                         CurrentWaitCountdownIndex++;
                     }
-                }
-            }
-
-            if (IsLaneActive)
-            {
-                if (CurrentTime >= LaneExpireTime)
-                {
-                    // Lane forgiveness window expired, disable all lane behavior
-                    RequiredLaneNote = -1;
-                    NextTrillNote = -1;
-                    YargLogger.LogFormatTrace("Lane behavior turned off at {0}", CurrentTime);
                 }
             }
         }
@@ -518,12 +509,12 @@ namespace YARG.Core.Engine
                 AdvanceToNextNote(note);
             }
 
-            if (!LanesExist || !note.IsLane || !BaseParameters.EnableLanes)
+            if ((!LanesExist && !note.IsLaneEnd) || !note.IsLane || !BaseParameters.EnableLanes)
             {
                 return;
             }
 
-            if (note.IsLaneStart || note.Time > LaneExpireTime)
+            if (note.IsLaneStart)
             {
                 YargLogger.LogFormatTrace("Starting lane behavior at time {0}. ", CurrentTime);
 
@@ -540,23 +531,22 @@ namespace YARG.Core.Engine
                 }
 
                 // Future updates during this lane will be handled on SubmitLaneNote inputs
-                UpdateLaneExpireTime();
+                UpdateLaneAutohitExpireTime();
             }
             else if (note.IsLaneEnd)
             {
-                // Lane ends with this note, continue overstrum leniency while transitioning out of this lane
                 YargLogger.LogFormatTrace("Lane ending at {0}", CurrentTime);
-
-                UpdateLaneExpireTime();
+                RequiredLaneNote = -1;
+                NextTrillNote = -1;
             }
 
             YargLogger.LogFormatTrace("Lane note hit at {0}", CurrentTime);
         }
 
         // Intercept a missed note while a lane phrase is active
-        protected bool HitNoteFromLane(TNoteType note)
+        protected bool AutohitNoteFromLane(TNoteType note)
         {
-            if (note.Time > LaneExpireTime)
+            if (note.Time > LaneAutohitExpireTime)
             {
                 return false;
             }
@@ -565,7 +555,7 @@ namespace YARG.Core.Engine
             {
                 if (note.IsLaneStart)
                 {
-                    // The leniency window at the end of the previous lane overlaps with the start of this one
+                    // The autohit window at the end of the previous lane overlaps with the start of this one
                     // The first note in a lane must be manually hit in order to count
                     return false;
                 }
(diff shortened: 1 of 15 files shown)


YARG.Core/Engine/Drums/DrumsEngine.cs
```diff
@@ -94,10 +94,28 @@ public virtual void Overhit()
                 return;
             }
 
-            // Prevent overstrum too close to the expiration of lane behavior
-            if (!IsLaneActive && CurrentTime - LaneExpireTime < LANE_END_LENIENCY)
+            // Prevent overhit too close to the beginning of a lane that accepts the overhit
```
wyrdough: I'm not really sure how I feel about giving overhit protection *before* the lane starts. I think it needs further discussion for me to really be comfortable with that.

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -162,6 +156,31 @@ protected virtual void Overstrum()
                 Codas[CurrentCodaIndex].Overhit();
             }
 
+            // Prevent overstrum too close to the beginning of a lane. Unlike the Keys and Drums engines,
+            // don't enforce the correct fretting; the player has the flexibility to adjust their fretting hand
+            if (
```
wyrdough: Just putting this here as a reminder in case we decide not to have pre-lane overstrum/overhit leniency.

YARG.Core/Engine/Keys/KeysEngine.cs
```diff
@@ -134,6 +134,31 @@ protected virtual void Overhit(int key)
                 return;
             }
 
+            // Prevent overhit too close to the beginning of a lane that accepts the overhit
+            if (
+                !IsLaneActive && // Not in a lane
```
wyrdough: Just putting this here as a reminder in case we decide not to have pre-lane overstrum/overhit leniency.

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -653,10 +655,30 @@ protected virtual bool ActiveLaneIncludesNote(int inputNote)
...
+
+        protected bool LaneThatJustEndedIncludesNote(int inputNote, TNoteType laneEnd)
+        {
+            var inputMask = 1 << inputNote;
+
+            var requiredLaneNote = laneEnd.LaneNote;
+            var nextTrillNote = laneEnd.IsTremolo ? -1 : laneEnd.PreviousNote!.LaneNote;
+
+            return inputMask == requiredLaneNote || (nextTrillNote != -1 && inputMask == nextTrillNote);
+        }
```
wyrdough: Why are these separate and not just a single `LaneIncludesNote`? You can decide whether you're checking `NextNote` or `PreviousNote` based on whether the passed note is a lane start or lane end note.

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -162,6 +156,31 @@ protected virtual void Overstrum()
...
+            )
+            {
+                YargLogger.LogFormatTrace("Overhit prevented by lane start leniency at {0}", CurrentTime);
+                return;
+            }
+
+            // Prevent overstrum too soon after the end of a lane. Unlike the Keys and Drums engines, don't
+            // enforce the correct fretting; the player has the flexibility to adjust their fretting hand
+            if (
+                !IsLaneActive && // Not in a lane
```
wyrdough: I think this (and the corresponding pre-lane overstrum/overhit check) would be more readable as a method called something like `IsInLeniencyWindow` or similar so we'd just be saying `if(IsInLeniencyWindow())`. It looks like you should be able to define that in `BaseEngine.Generic`.

wyrdough (changes requested):
Need to do some testing to verify functionality, but these are the things that jumped out at me as areas for improvement/discussion.

I do think that, in general, this is a better approach than what we were previously doing, so thanks for taking this on!

YARG.Core/Engine/Drums/DrumsEngine.cs
Frickitickitavi: I think it's an important part of keeping consecutive lanes loose and comfortable for players. Back-to-back lanes usually represent loose transitions, and I want the player to have wiggle room (in both directions) as to when exactly they switch from one lane input to the next.

image

Another part of my motivation here is how we currently render open lanes (which will also apply to wildcard and kick lanes, eventually):

image


The visual lane extends much earlier than the actual effect of the lane, and in testing I constantly find myself strumming early and overhitting. Regular lanes don't have this visual quirk; I assume we give open lanes this effect to keep them visible under the bar. If we can change that and make the open lane visuals look less misleading, that's cool too, although the first point about transitions still stands.

YARG.Core/Engine/BaseEngine.Generic.cs
Frickitickitavi: That's fair - I'll do that

YARG.Core/Engine/Guitar/GuitarEngine.cs
Frickitickitavi: Agreed, will do

Frickitickitavi: Done, although I ended up creating two overloaded `IsInLeniencyWindow` methods instead of just one. One is parameterless, used for Guitar, and the other takes the input note, which Drums and Keys care about for checking leniency windows.

YARG.Core/Engine/BaseEngine.cs
jstnlef: Before this line I think you should use 
`var update = _scheduledUpdates[0];` and use that  in the subsequent calls prior to line 300. This would turn updateTime into `double updateTime = update.Time;` and other calls into `update.Reason`

jstnlef, line 279: There should be a `continue;` here. If not, updateTime and _scheduledUpdates[0].Reason no longer point to the same event

jstnlef: This should be removed.

jstnlef, line 310: I recommend the following be added
```
private void GenerateAndSortQueuedUpdates(double nextTime)
        {
            GenerateQueuedUpdates(nextTime);
            _scheduledUpdates.Sort((x, y) => x.Time.CompareTo(y.Time));
        }
```
and used in RunQueuedUpdates such that we don't accidentally forget to sort the updates post GenerateQueuedUpdates calls

Frickitickitavi: Just the `var previousTime = CurrentTime;` line, right?

jstnlef: yep. Just the unused var

Frickitickitavi: Done - please double-check that I updated the correct locations

Frickitickitavi: Done

Frickitickitavi: Done

Frickitickitavi: Done

jstnlef: This should be updated to call the new GenerateAndSortQueuedUpdates as well. Sorry about missing this the first time.

jstnlef: This comment should be removed too because I don't know what this is associated with.

jstnlef (changes requested):
Just one more small change and then it looks good. Thanks!

YARG.Core/Engine/BaseEngine.cs
Frickitickitavi: No worries - done. Thanks!

Frickitickitavi: Yeah I was wondering about that too, since this uses `while` instead of `for` - removed

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -162,6 +156,15 @@ protected virtual void Overstrum()
                 Codas[CurrentCodaIndex].Overhit();
             }
 
+            // Prevent overstrum too close to a lane. Unlike the Keys and Drums engines, use the lenient
```
jstnlef: Should this happen before the "IsCodaActive" check? I suspect it probably should such that it bails early

YARG.Core/Engine/BaseEngine.cs
```diff
@@ -257,10 +258,7 @@ private void ProcessInputs(double time)
 
         private void RunQueuedUpdates(double time)
         {
-            // 'for' is used here to prevent enumeration exceptions,
-            // the list of scheduled updates will be modified by the updates we're running
-
-            GenerateQueuedUpdates(time);
+            GenerateAndSortQueuedUpdates(time);
             _scheduledUpdates.Sort((x, y) => x.Time.CompareTo(y.Time));
```
jstnlef: This should be removed or we're sorting twice ;)

YARG.Core/Engine/BaseEngine.Generic.cs
```diff
@@ -653,10 +655,86 @@ protected virtual bool ActiveLaneIncludesNote(int inputNote)
             return false;
         }
 
-        protected void UpdateLaneExpireTime()
+        // Lenient version, which only cares about proximity to any lane, not the contents of the lane.
+        // Used by Guitar engines to allow for fretting flexibility during transitions
+        protected bool IsInLaneLeniencyWindow()
```
jstnlef: If this is only to be used with Guitar engines, shouldn't it live in `GuitarEngine.cs`?

YARG.Core/Engine/BaseEngine.cs
Frickitickitavi: Oh right, duh lol

YARG.Core/Engine/BaseEngine.Generic.cs
Frickitickitavi: Yeah fair - I was thinking of it as a general lenient version which just happens to _currently_ be only used by guitar (as opposed to an intrinsically guitar-only thing), but I guess that kind of falls under YAGNI and I should put it where it currently belongs

jstnlef: We can always move it later if it's needed somewhere else :)

YARG.Core/Engine/Guitar/GuitarEngine.cs
Frickitickitavi: Yeah, good call. Technically `CodaHasStarted` is the one that matters, not `IsCodaActive`, but it makes sense to put this before both of them to keep the coda checks together.

jstnlef (approved):

wyrdough (approved):

PR #386, opened 2026-05-27
fix highway lowering early bug in some charts

4dispel:
if the last note of a chart is a chord with different length notes currently we lower the highway after the parent note in the chord ends.
this leads to for example any GRYBO non-sustain note on top of an open sustain lowering the highway immediately, because an open note in a chord is reliably parsed as a child.

To Fix this, this PR adds a new method to replace GetEndTime for Lists of Notes that iterates over the children of the last note to get the actual end of the chord.


4dispel:
changed to taking the maximum of all notes instead of just the last chord to account for extended sustains as well

wyrdough (approved):

PR #389, opened 2026-05-31
Fix lyric parsing on the same tick as phrase start/end (+tests)

Supersonicboss1:
This matches the behavior from CH. I also wrote some tests for this, though I haven't tested this with MIDI files.

The behavior that should be expected for some event is:
1. On the same tick as a phrase_start, it should go into that phrase.
2. Otherwise, if it's on the same tick as a phrase_end, it should go into that phrase.
3. If it's outside of a phrase, ignore the event.
4. If it's in a phrase, put the event in the phrase.

Also, changed `FlushPendingEvents` to actually *flush* the events instead of just adding them.


jstnlef (approved):
This seems alright to me 👍. Love the tests.

PR #391, opened 2026-06-04
Implement postproc lane parsing

Frickitickitavi:
Completely reimplements how we parse trill and tremolo lanes. The previous logic operated on `MoonNote` data, which made it very difficult to properly validate and sanitize lanes, since they depend heavily on relationships between notes and the `MoonNote` parser has no sense of state.

`MoonSongLoader::LoadDifficulty` now takes one additional delegate, `FinalPassDelegate`. This delegate operates on an entire `InstrumentDifficulty<TNote>` - the thing we previously would have returned - and allows for final-pass logic that depends on fully-constructed `Note` and `Phrase` data rather than `MoonNote` data or incomplete `Note` lists. It's currently used only for lane parsing, but in theory it could be used for anything else that similarly benefits from preprocessed data.

We now heavily validate and sanitize lanes as we parse them. If a lane is malformed, we terminate it at the first invalid note, or reject the lane entirely if it's invalidated before the lane can be established (before 2 valid tremolo notes or 3 valid trill notes).

Guitar and ProKeys effectively share the same logic for trills, but because `NoteMask` is independently defined in each of their note classes rather than being inherited from `Note`, I still had to do a bit of code duplication. We can probably get clever with some kind of `GetNoteMaskDelegate` or something if you want - this is just a first draft.

The Drums loader has very different rules for lane validation, especially for tremolos. I've heavily commented all of the new logic, so hopefully it's easy enough to follow. Again, this is a first draft, and I'm sure there are some places where we could clean up the flow/readability of things.

### DIFF
--- a/YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Drums.cs
+++ b/YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Drums.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using MoonscraperChartEditor.Song;
 using MoonscraperChartEditor.Song.IO;
+using YARG.Core.Extensions;
+using YARG.Core.Logging;
 using YARG.Core.Parsing;
 using static MoonscraperChartEditor.Song.MoonNote;
 
@@ -36,12 +38,12 @@ namespace YARG.Core.Chart
 
             var difficulties = new Dictionary<Difficulty, InstrumentDifficulty<DrumNote>>()
             {
-                { Difficulty.Beginner, LoadDifficulty(instrument, Difficulty.Beginner, beginnerNoteDelegate, HandleTextEvent) },
-                { Difficulty.Easy, LoadDifficulty(instrument, Difficulty.Easy, createNote, HandleTextEvent) },
-                { Difficulty.Medium, LoadDifficulty(instrument, Difficulty.Medium, createNote, HandleTextEvent) },
-                { Difficulty.Hard, LoadDifficulty(instrument, Difficulty.Hard, createNote, HandleTextEvent) },
-                { Difficulty.Expert, LoadDifficulty(instrument, Difficulty.Expert, createNote, HandleTextEvent) },
-                { Difficulty.ExpertPlus, LoadDifficulty(instrument, Difficulty.ExpertPlus, createNote, HandleTextEvent) },
+                { Difficulty.Beginner, LoadDifficulty(instrument, Difficulty.Beginner, beginnerNoteDelegate, HandleTextEvent) }, // No lanes on Beginner, so no final pass
+                { Difficulty.Easy, LoadDifficulty(instrument, Difficulty.Easy, createNote, HandleTextEvent, finalPassDelegate: DrumsFinalPass) }, // Drum lanes on Easy are possible in .chart, so we do need the final pass
+                { Difficulty.Medium, LoadDifficulty(instrument, Difficulty.Medium, createNote, HandleTextEvent, finalPassDelegate: DrumsFinalPass) }, // Drum lanes on Medium are possible in .chart, so we do need the final pass
+                { Difficulty.Hard, LoadDifficulty(instrument, Difficulty.Hard, createNote, HandleTextEvent, finalPassDelegate: DrumsFinalPass) },
+                { Difficulty.Expert, LoadDifficulty(instrument, Difficulty.Expert, createNote, HandleTextEvent, finalPassDelegate: DrumsFinalPass) },
+                { Difficulty.ExpertPlus, LoadDifficulty(instrument, Difficulty.ExpertPlus, createNote, HandleTextEvent, finalPassDelegate: DrumsFinalPass) },
             };
 
             foreach (var difficulty in difficulties)
@@ -121,7 +123,6 @@ namespace YARG.Core.Chart
             var noteType = GetDrumNoteType(moonNote);
 
             var generalFlags = GetGeneralFlags(moonNote, currentPhrases, IsEligibleForDrumTrill);
-            generalFlags = ModifyDrumLaneFlags(moonNote, currentPhrases, generalFlags, notes);
 
             var drumFlags = GetDrumNoteFlags(moonNote, currentPhrases);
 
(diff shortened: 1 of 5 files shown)


YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Drums.cs
```diff
@@ -437,130 +437,396 @@ private InstrumentDifficulty<DrumNote> LoadFromEliteDrumsDownchartDifficulty(Ins
...
-                return validTremoloNotes;
+            if (pad2 is null)
+            {
+                // We didn't find a valid second pad (presumably because the phrasae was nothing but kicks after the first hit), so this is not a valid trill
+                return null;
             }
+
+            // We have two valid trill notes. We just need to make sure that the next non-lone-kick hit matches the first, and we have our pads
+            noteRef = noteRef!.NextNote;
+            while (noteRef is not null)
```
wyrdough: It's 3 in the morning so I may not be all there, but don't we want to make sure that the candidate note here is actually in the phrase? In other words, shouldn't we be checking for the `LaneEnd` flag and refusing to create a lane if the note we're evaluating has that flag and doesn't resolve the lane? (and isn't a kick whose single child resolves the lane, I guess?)

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Guitar.cs
```diff
@@ -221,5 +221,191 @@ private bool IsWithinCoda(Phrase phrase)
...
+                switch (phrase.Type)
+                {
+                    case PhraseType.TremoloLane:
+                        laneNotes = GetGuitarTremoloNotes(notesInPhrase);
+                        break;
+                    case PhraseType.TrillLane:
+                        laneNotes = GetGuitarTrillNotes(notesInPhrase);
+                        break;
+                    default:
+                        throw new ArgumentOutOfRangeException("Unreachable.");
```
wyrdough: The first parameter to `ArgumentOutOfRangeException` is supposed to be the name of the parameter.


```suggestion
                        throw new ArgumentOutOfRangeException(nameof(phrase.Type), phrase.Type, "Invalid phrase type for Guitar lane");
```

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Guitar.cs
```diff
@@ -221,5 +221,191 @@ private bool IsWithinCoda(Phrase phrase)
...
+                )
+                {
+                    note.ActivateFlag(NoteFlags.Trill);
+                    trillNotes.Add(note);
+                }
+                else
+                {
+                    // We've stopped properly alternating; terminate the trill
+                    break;
+                }
```
wyrdough: I think this would be slightly more readable like this:


```suggestion
            var mask = notesInPhrase[0].NoteMask;
            var otherMask = notesInPhrase[1].NoteMask;

            for (var i = 3; i < notesInPhrase.Count; i++)
            {
                var note = notesInPhrase[i];

                if (note.NoteMask == otherMask)
                {
                    note.ActivateFlag(NoteFlags.Trill);
                    trillNotes.Add(note);
                }
                else
                {
                    // We've stopped properly alternating; terminate the trill
                    break;
                }

                (mask, otherMask) = (otherMask, mask);
            }
```

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.ProKeys.cs
```diff
@@ -66,5 +66,121 @@ private ProKeysNoteFlags GetProKeysNoteFlags(MoonNote moonNote, Dictionary<MoonP
...
+                {
+                    note.ActivateFlag(NoteFlags.Trill);
+                    trillNotes.Add(note);
+                }
+                else
+                {
+                    // We've stopped properly alternating; terminate the trill
+                    break;
+                }
+            }
```
wyrdough: See the corresponding note about `MoonSongLoader.Guitar`...

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Drums.cs
```diff
@@ -437,130 +437,396 @@ private InstrumentDifficulty<DrumNote> LoadFromEliteDrumsDownchartDifficulty(Ins
...
+                        if (candidatePads[0] == child.Pad)
+                        {
+                            // The pad reoccurred, so this is a valid tremolo of that pad
+                            return candidatePads[0];
+                        }
                     }
 
-                    int thisNote = noteRef.rawNote;
+                    return null; // The pad never reoccurred, so this is not a valid tremolo
+                }
```
wyrdough, line 581: You're not actually checking subsequent notes, only the children of `notesInPhrase[i]` since you're immediately returning null if the inner `foreach` hasn't returned. If you don't need to check subsequent notes, get rid of the outer loop. Otherwise, move the `return null` to just after the for loop. No need to create and check a flag since the inner `foreach` will have returned if it was a valid tremolo.


```suggestion
                for (var i = 1; i < notesInPhrase.Count; i++)
                {
                    foreach (var child in notesInPhrase[i].AllNotes)
                    {
                        if (otherPadOfSameColor is not null && (int)otherPadOfSameColor == child.Pad)
                        {
                            // The other pad of the same color reoccurred before the candidate did, so this is not a valid tremolo
                            return null;
                        }

                        if (candidatePads[0] == child.Pad)
                        {
                            // The pad reoccurred, so this is a valid tremolo of that pad
                            return candidatePads[0];
                        }
                    }
                }
                return null; // The pad never reoccurred, so this is not a valid tremolo
```

wyrdough (changes requested):
Thanks for taking this on. Needs some minor changes, but generally seems like a good approach to me!

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Drums.cs
Frickitickitavi: Ah yup - I caught myself doing the same thing in one of the other loaders and fixed it, but I guess I let it slip through here.

Checking for `LaneEnd` doesn't work, because we haven't set that yet. But what I did in the other cases will work here too, which is to do `for (var i = 2; i < notesInPhrase.Count; i++)` instead of of `while (noteRef is not null)` with a `noteRef = noteRef.NextNote` inside of it. That way, we fall out of the loop when we're out of in-phrase notes rather than continuing through the entire chart

(we start at `i=2` because we've already vetted the first two notes of the trill)

YARG.Core/Chart/Loaders/MoonSong/MoonSongLoader.Drums.cs
Frickitickitavi: We do want to check subsequent notes, so doing the latter (moving the `null` return outside the outer loop)

Frickitickitavi:
All comments addressed - will run some tests to make sure everything's still in working order

wyrdough (approved):

PR #394, opened 2026-06-09
Add OnStarPowerGain event for gain SFX

Supersonicboss1:
Previously, this was being handled in YARG separately in the vocals and track players for notifications. Since I needed it for SFX anyway, and that seemed at odds with the other star power events, I made it an event instead.

### DIFF
--- a/YARG.Core/Engine/BaseEngine.cs
+++ b/YARG.Core/Engine/BaseEngine.cs
@@ -20,7 +20,7 @@ namespace YARG.Core.Engine
         protected const int SUSTAIN_BURST_FRACTION = 4;
 
         public delegate void StarPowerStatusEvent(bool active);
-        public delegate void StarPowerGainEvent();
+        public delegate void StarPowerReadyEvent();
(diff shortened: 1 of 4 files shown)


YARG.Core/Engine/BaseEngine.cs
```diff
@@ -29,6 +30,7 @@ public abstract class BaseEngine
 
         public delegate void UnisonBonusAwardedEvent();
         public StarPowerStatusEvent? OnStarPowerStatus;
+        public StarPowerGainEvent?   OnStarPowerGain;
```
wyrdough: I think the name `OnStarPowerGain` is slightly confusing since it only actually fires when the player has a usable amount of SP, not literally whenever SP increases.

wyrdough (changes requested):

YARG.Core/Engine/BaseEngine.cs
Supersonicboss1: Okay, renamed to OnStarPowerReady (I also renamed the SFX file in the other PR)

PR #397, opened 2026-06-13
Separate saved and effective vocal modifiers

XaiaX:
### Summary

Splits `YargProfile`'s modifier state into a *saved selection* (what the player explicitly chose; what profile JSON persists) and the *effective* value (what gameplay reads), mirroring the `DifficultyFallback` pattern. Adds `ApplySessionModifiers(YargProfile source)`, which sets only the effective value — letting a caller impose another profile's modifiers for the current session without erasing the profile's own saved selection.

### Why

At song launch the difficulty select menu enforces uniform vocal modifiers by copying the selecting player's modifiers onto every other vocal player (`CopyModifiers`). Uniformity is the intended semantics — only one vocal player is allowed to edit modifiers per song — but the copy writes the same field that profile JSON persists, so it **permanently overwrites** the other players' saved selections. Repro: give a second vocal profile `NoVocalPercussion`, play one song where the first profile (no modifiers) is the selector — the second profile's saved modifier is now reset to None, on disk.

### Design

- `_savedModifiers` (new private field) is serialized under the existing `"CurrentModifiers"` JSON property name → **existing profiles load unchanged**, and saves from this version load in older builds.
- `CurrentModifiers` (the public property) becomes the effective value, `[JsonIgnore]`d.
- Explicit edits (`AddSingleModifier`/`RemoveModifiers`) and `CopyModifiers` write **both** — so every existing caller behaves exactly as before.
- `ApplySessionModifiers` writes only the effective value.
- `RestoreSavedModifiers` re-seeds the effective value from the saved selection, for discarding a prior session's imposed modifiers. Needed because profiles are long-lived in memory: without it, modifiers applied via `ApplySessionModifiers` linger on the effective value and reappear the next time the value is read, even though the persisted (saved) selection is intact.
- Both deserialization paths seed both fields (JSON via the existing `[OnDeserialized]` hook; the replay binary constructor inline).
- Replay serialization still writes the **effective** `CurrentModifiers` — the modifiers gameplay actually ran — so replays and `ReplayAnalyzer` are unaffected under either caller.

### Changes

| File | Change |
|------|--------|
| `YARG.Core/Game/YargProfile.cs` | `_savedModifiers` field (serialized as `"CurrentModifiers"`), effective `CurrentModifiers`, `ApplySessionModifiers`, `RestoreSavedModifiers`; both ctors and `[OnDeserialized]` seed both fields |
| `YARG.Core.UnitTests/Game/YargProfileModifierTests.cs` | New: session-apply changes effective state; session-apply doesn't survive JSON round-trip; restore discards session modifiers / is idempotent for an untouched profile; `CopyModifiers` still persists (inertness guarantee); explicit edits persist; deserialization seeds effective; legacy JSON property-name compatibility |

### Notes

- **No observable behavior change until something calls `ApplySessionModifiers`** (the Unity PR). Until then the two fields always hold the same value — verified by unit test (`CopyModifiersStillPersists`) and manually in the Unity editor (modifier selection, gameplay application, persistence, and the clobber bug itself all behave identically).
- Profile JSON format and replay binary format are unchanged.
- Safe to merge before the Unity PR, and required for it to compile.


polson:
Resolve conflicts and I'll get this merged

polson (approved):

PR #404, opened 2026-06-21
fix(engine): suppress end-tick BRE finale through the whole coda

XaiaX:
## Summary

Ties the BRE-note suppression gates to `CodaHasStarted` instead of `IsCodaActive` across all four engines: `DrumsEngine`, `FiveLaneKeysEngine`, `ProKeysEngine` (`HitNote` + `MissNote`) and `GuitarEngine.HitNote`, matching what `GuitarEngine.MissNote` already does (`9b8802fa4`). Also normalizes the drums/keys miss branches to `SetHitState(true, true)`.

## Why

A note charted exactly on the BRE-end tick is `IsBigRockEnding` and hidden under the BRE lanes, so it must be auto-suppressed (freestyle) so the bonus triggers regardless of whether or not it's hit. Instead it's treated as a normal must-hit note: missing it forfeits the BRE bonus, and hitting it (judged after `EndTime`) potentially drives `NotesMissed` **negative** on a full combo, because `TotalNotes` excludes BRE notes but the normal hit path still increments `NotesHit`.

`060ed95a` (`time >= EndTime` → `time > EndTime`) addresses the *hit-path* edge at the `EndTime` boundary, but the *miss path* is judged at the finale's back-end (`note.Time + backEnd`), strictly after `EndTime`, where `IsCodaActive` is false under either operator. So on current `master` the missed finale still falls through and the bug persists (in-game: `NotesMissed` goes to -2, bonus lost unless the invisible notes are hit).

## Root cause (short)

`IsCodaActive` deactivates at `EndTime`; a note is judged at its back-end (after `EndTime`) so the `IsCodaActive && note.IsBigRockEnding` is no longer true. `GuitarEngine.MissNote` uses `CodaHasStarted` (`9b8802fa4`) so it didn't trigger this problem.

## Fix / Design

Change the gates to `CodaHasStarted && note.IsBigRockEnding` (what `GuitarEngine` uses). `CodaHasStarted` stays true until `EndCoda()` fires (when the finale resolves), so it covers the finale's whole judgment window. The change leaves `IsCodaActive` (and its window) untouched. It still governs fret/key freestyle input, overhit logic, etc.

## Changes

| File | Change |
|------|--------|
| `YARG.Core/Engine/Drums/DrumsEngine.cs` | `HitNote`+`MissNote` gates → `CodaHasStarted`; miss branch `SetHitState(true, true)` |
| `YARG.Core/Engine/Guitar/GuitarEngine.cs` | `HitNote` gate → `CodaHasStarted` (`MissNote` already correct) |
| `YARG.Core/Engine/Keys/FiveLaneKeys/FiveLaneKeysEngine.cs` | `HitNote`+`MissNote` gates → `CodaHasStarted`; miss branch `SetHitState(true, true)` |
| `YARG.Core/Engine/Keys/ProKeys/Engines/ProKeysEngine.cs` | `HitNote`+`MissNote` gates → `CodaHasStarted`; miss branch `SetHitState(true, true)` |
| `YARG.Core.UnitTests/Engine/BreSoftlockTests.cs` | +4 `EndTickFinalChord` tests: drums/five-lane-keys/pro-keys repros + a guitar regression guard, sharing a `RunWithoutHittingFinale` helper |

## Testing done

- **Unit (GREEN):** `dotnet test YARG.Core.UnitTests` → **419 passed, 2 skipped** (`FullScan`/`QuickScan`), 0 failed.
- **RED/GREEN (run against `060ed95a`):** before the fix, the 4 `EndTickFinalChord` tests report **Failed: 3, Passed: 1** (drums / five-lane-keys / pro-keys fail, `CodaSuccess = false`); guitar passes (its `MissNote` was already `CodaHasStarted`). After the fix, all 4 pass.
- **In-game (2026-06-21):** `dirtydeeds_live`, FourLane Expert, BRE phrase `288000..296160`, finale `[Yellow+Blue]` at tick `296160`. On base `master`: missing the hidden finale forfeits the bonus and drives `NotesMissed` to -2. After fix, BRE finale works as expected.

## Notes

- **Builds on `060ed95a`** (wyrdough's `IsCodaActive` inclusive-end change). The two coexist; the `CodaHasStarted` gate is the load-bearing part for the miss path. If preferred, `060ed95a`'s one-line change could be reverted, but leaving it is harmless.
- The 7 gate sites are the complete set (`grep "IsCodaActive && note.IsBigRockEnding"` returns nothing after this PR); no `Yarg*` subclass re-gates differently.

### DIFF
--- a/YARG.Core.UnitTests/Engine/BreSoftlockTests.cs
+++ b/YARG.Core.UnitTests/Engine/BreSoftlockTests.cs
@@ -7,6 +7,8 @@ using YARG.Core.Engine.Drums;
 using YARG.Core.Engine.Drums.Engines;
 using YARG.Core.Engine.Guitar;
 using YARG.Core.Engine.Guitar.Engines;
+using YARG.Core.Engine.Keys;
+using YARG.Core.Engine.Keys.Engines;
 using YARG.Core.Game;
 using YARG.Core.Input;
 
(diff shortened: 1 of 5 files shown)


wyrdough (approved):
Good enough for now... 

I do want to get rid of `IsCodaActive` and `CodaHasStarted` at some point, though. Auto-hit can depend solely on whether the note is flagged with `IsBigRockEnding` and we can add a `CodaStart` flag during chart parsing to trigger `OnCodaStarted` when we advance to the first BRE note and use the existing `CodaEnd` flag to trigger `OnCodaEnded` once we advance past that note.

PR #405, opened 2026-06-26
Add vocals censorship chart feature (core)

Supersonicboss1:
As described here: https://docs.google.com/document/d/1L5yDejAOpbSnDvicgh6Id0aIeMewORVBQRStErAnogw/edit?usp=sharing

This encompasses three parts, each of which are technically independent from each other, but all of which depend on the "Censor Mature Content" setting being enabled in YARG:
1. Stems - Depends on [stem]_clean or [stem]_explicit files being present. If censor mature content is disabled, loads the normal vocal stem, and if present, the _explicit file. If the setting is enabled, attempts to load _clean, then if it does not find that for a particular stem, loads the normal variant.
2. Content Rating - If a song is rated as "Mature" or "Sensitive Content", and `clean_vocals = 1` is found in the ini, drop the rating to "Supervision Required".
3. Chart - If the `[CENSORSHIP_MARKERS]` text event is found in that vocals part, anywhere where MIDI note 104 is covering a vocal note, remove the note, and replace the entire word with a dash. The state of this setting is also stored in replays, to ensure correct playback.

While I am relatively happy with the stem and content rating implementations, I am unsure about the chart stuff. The way I have it implemented here is as a public post-processing step in SongChart, which applies the transformation. Not sure if this is better, or if passing it further up the chain to MoonSongLoader (or handling it in GameManager) is a better idea. Furthermore, because of the way this works, it requires looking at vocal notes in MoonSongLoader's lyric parsing which doesn't deal with flags at all, making it a bit less clean.

Also, sorry about the random whitespace changes, I have no idea how to make Rider stop doing that.

Depends on YARC-Official/YARG#1529


wyrdough:
I have not done a full review, but a quick read-through of the diff did not show any obvious glaring architectural issues, so I don't think any massive redesign is going to be required before merging this.

Thanks for handling this, I know there are some team members who are quite interested.

YARG.Core/Chart/Notes/VocalNote.cs
```diff
@@ -119,6 +125,7 @@ public VocalNote(NoteFlags noteFlags, bool isPercussionPhrase,
 
             TotalTimeLength = timeLength;
             TotalTickLength = tickLength;
+            IsCensorable = false; // Percussion phrase should never be censored
```
wyrdough: I don't see where we're actually checking that this is a percussion phrase?

YARG.Core/Song/Entries/Ini/SongEntry.UnpackedIni.cs
```diff
@@ -40,28 +40,48 @@ internal override void Serialize(MemoryStream stream, CacheWriteIndices node)
...
             foreach (var stem in IniAudio.SupportedStems)
             {
                 var stemEnum = AudioHelpers.SupportedStems[stem];
-                if (ignoreStems.Contains(stemEnum))
+
+                if (ignoreStems.Contains(stemEnum) || addedCleanStems.Contains(stemEnum))
+                {
                     continue;
+                }
+                TryLoadStem(stem, stemEnum, subFiles, mixer);
```
wyrdough, line 71: I may well be misunderstanding because I'm just looking at the diff on GitHub, but it looks to me like if censoring is not enabled we could potentially be loading two stems since you're calling `TryLoadStem` here and then in the `!enableCensoring` path below you're also potentially loading an explicit stem.

YARG.Core/Song/Entries/SongEntry.cs
```diff
@@ -109,8 +109,6 @@ public abstract partial class SongEntry
 
         public int PlaylistTrack => _metadata.PlaylistTrack;
 
-        public SongRating SongRating => _metadata.SongRating;
```
wyrdough, line 112: Shouldn't this have changed to `public SongRating SongRating => GetSongRating();` instead of being deleted outright?

wyrdough (changes requested):
This is what I can see from looking at the diff alone. Feel free to push back, because I'm definitely missing context.

YARG.Core/Song/Entries/Ini/SongEntry.UnpackedIni.cs
Supersonicboss1: This is the intended behavior - the _explicit stems only contain the explicit parts of the track (e.g. swear words and the like) so they need to be loaded in addition to the normal stem.

YARG.Core/Chart/Notes/VocalNote.cs
Supersonicboss1: My bad, I didn't realise what this constructor did. I still think this should always be false, since this is the constructor for a vocal/percussion phrase, and that should never be censored, but the comment is wrong.

YARG.Core/Song/Entries/SongEntry.cs
Supersonicboss1: I wasn't sure if I should enforce an explicit contract with any code that gets SongRating - essentially, should they be required to consider explicitly whether to return that with censorship enabled or not?

YARG.Core/Chart/Notes/VocalNote.cs
wyrdough: Given that there is an explicit check for whether or not `isPercussionPhrase` is set, probably best not to assume that it is or will only ever be used for percussion phrases.

Supersonicboss1: Sorry, I don't think I was clear in my reply (though the new commit helps explain it a bit better). According to the comment on this constructor, it should only be used for either a lyric phrase, or percussion phrase. Regardless of which it is, we do not censor entire phrases, and should not censor the parent phrase "notes", only the vocal notes themselves.

YARG.Core/Song/Entries/SongEntry.cs
Hubbubble1: Here's what frick and I figure: since the instance a SongEntry pertains to is fixed to a particular state of a song being censored or uncensored and won't change, we should use the derived value like wyrdough suggests to avoid having to compute whether it is SR or not. Also worth mentioning if this needs to be changed in the future, it wouldn't be a big deal to adjust this part

YARG.Core/Song/Entries/SongEntry.cs
```diff
@@ -376,6 +374,7 @@ internal virtual void Serialize(MemoryStream stream, CacheWriteIndices node)
             stream.Write(_metadata.SongLength, Endianness.Little);
             stream.Write(_metadata.SongOffset, Endianness.Little);
             stream.Write((int)_metadata.SongRating, Endianness.Little);
+            stream.Write(_metadata.CleanVocals);
```
polson, line 377: I think we need to update `CACHE_VERSION` inside `CacheHandler` if this changes?

YARG.Core/Song/Entries/Ini/SongEntry.UnpackedIni.cs
```diff
@@ -40,28 +40,48 @@ internal override void Serialize(MemoryStream stream, CacheWriteIndices node)
                 YargLogger.LogError("Failed to create mixer!");
                 return null;
             }
-
+            var addedCleanStems = new HashSet<SongStem>();
```
polson, line 43: This might not work for like vocals_1 and vocals_2 if the song has it, since it is keying off of the song stem, and we collapse those to a single song stem

polson (changes requested):

YARG.Core/Song/Entries/Ini/SongEntry.UnpackedIni.cs
Supersonicboss1: I'm aware of this, but not sure if it's worth fixing for the extra complexity it adds, considering it's only an issue in the case where:
1. A song has multiple vocals stems.
2. Some, but not all of these stems are clean/explicit, and some are not.

To fix this would probably involve doing either manually stripping the _clean or _explicit suffixes from files to relate them to their normal counterparts, or a dictionary map for these, but either way feels slightly unclean.

YARG.Core/Song/Entries/SongEntry.cs
Supersonicboss1: Okay, looked into this a bit, and while it's doable, it's a bit of a pain. The options are:
1. Still pass in whether censoring is enabled on every call, but if censoring has ever been enabled on this instance, skip the checks and just give it the censored version. This is likely to cause weirdness with state if anyone forgets to do this, and I don't really like it because SongEntry doesn't really have state to my knowledge.
2. Pass whether censoring is enabled through the song cache code and subclasses so that it's set on initialisation. This works, is probably best for performance, but is much more invasive since it would involve passing a boolean parameter through practically all the song scanning code, and would require a rescan when changing the setting.
3. Precalculate what the "effective" rating would be when censoring is enabled - still pass it in as a parameter on every call, but just return that precalculated value if censoring was enabled. Worse for performance than 2, though probably better than now. I could also make this behave like 1 if no argument is provided (e.g. the derived value).

YARG.Core/Song/Entries/SongEntry.cs
Supersonicboss1: true, done

polson (approved):

YARG.Core/Replays/ReplayInfo.cs
```diff
@@ -128,6 +135,7 @@ public void Serialize(BinaryWriter writer)
             writer.Write((byte) BandStars);
 
             writer.Write(Pauses.Length);
+            writer.Write(CensorshipEnabled);
```
polson: Need to make sure your reads and writes happen in the same order or you will read the wrong data

Supersonicboss1: Good catch, fixed

PR #406, opened 2026-06-30
Add SongIniWriter for writing song.ini delay offset

asimard1:
Adds SongIniWriter.AddSongOffset(iniPath, offset), which reads an existing song.ini, adds offset (ms) to the existing delay value if one exists, or inserts a new delay line under [song] if not.
Used by a companion PR in the main YARG repo to power an in-game offset calibration button on the score screen: YARC-Official/YARG#1535.


wyrdough:
Thanks for the PR!

Just to explain some back story that I don't think was really covered on Discord, we inherited a strong aversion to ever editing the songs in a user's library directly. The main reasons being that we expect that libraries will be shared between games, so we shouldn't modify songs in order to avoid causing issues for other games and also that if we don't even have the capability of writing to the user's library folders we can be assured that nothing in YARG can go haywire and corrupt someone's songs. We might manage to screw up our own song cache, but that won't affect the underlying library so it can be used for other games or the song cache can be deleted and recreated easily enough. If we screw up people's songs, that's not recoverable.

All that said, I am in favor of the underlying functionality of making it easier for the player to save an adjustment to the audio offset of the song. Wonky charts are not that uncommon, after all. My inclination would be to not save it in the .ini, though. We could either extend what is saved in the song cache to allow offsets to be saved there or perhaps write out a json file with offsets similar to how we save favorites and other user data. By not saving it in the song itself, we also get the side benefit of it being able to work with any supported song format, not just unpacked ini.

asimard1:
I totally understand the mentality, and I could change my other PR to make that work (writing to a json file and all).

Do you have any pointers on how to create the .json file in a way which goes along with the game's mentality (e.g. where the .json file should be located in the game files, how the keys should be stored, etc.)?

edit: I'll start by reading the code for the favorites to see how that is stored, and work from there.

PR #407, opened 2026-07-01
Add Tournament Engine

Ape:
See https://github.com/YARC-Official/YARG/pull/1539.


wyrdough (approved):
Assuming you got the ghost/accent note bonus and total sustain score correct, I see nothing wrong with this.

YARG.Core/Engine/BaseEngineParameters.cs
```diff
@@ -78,6 +96,21 @@ protected BaseEngineParameters(ref FixedArrayStream stream, int version)
             {
                 EnableLanes = stream.ReadBoolean();
             }
+
+            if (version >= 17)
```
polson: Current replay version is 17, so then I think we are trying to read stuff that doesn't exist in old replays and they probably will break

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -443,6 +443,45 @@ protected sealed override (int, int) CalculateChartScores()
...
+                maxScore += multiplier * pointsForSustain;
+
+                if (note.IsDisjoint)
+                {
+                    HashSet<uint> seenNoteTicks = new();
+                    foreach (var child in note.ChildNotes)
+                    {
+                        if (seenNoteTicks.Add(child.Tick))
+                        {
+                            combo++;
```
polson, line 472: Double check this is right

polson (changes requested):

polson:
I think it might be best if we got a testing framework in for scoring first before we do anything with this.  There's a lot of scoring logic that remains untested, tournament mode or not, and something like this could potentially get bonuses or max scoring wrong, as wyrdough mentioned

polson (changes requested):
One thing in replay version to fix

Ape:
Rebased on bumped replay version.

PR #411, opened 2026-07-09
Add CON update support for midi/ini songs that have a shortname tag (that can be created from the CON)

asimard1:
Adds song_update.mid support to ini-format songs, mirroring how CON packs already handle .mid updates. This is useful for people who want to use CON updates but still see an advantage in having extracted files (for different reasons).


asimard1:
This now works perfectly and all hash values fit except for a very few (2 songs on multiple thousands).

The only thing I'm not sure how to do: How do I hide YARG.Core code behind the settings defined in YARG?... I'll have to read some code and add that.

edit: Maybe I'll add it as a parameter to the scan? I'll look at this further tomorrow. Although, the CON updates for regular CONs aren't hidden behind a toggle, so is it necessary to add the option?

Although maybe I should add serious support for the other parts of CON updates (i.e. the album art, audio, new song infos, etc.).

asimard1:
Last thing TODO: add support for the dta updates

asimard1:
Finally done and cleaned up, tested in game as well.
The only negative thing about this is the fact that scanning takes longer.

YARG.Core/Song/Cache/CacheHandler.cs
```diff
@@ -304,11 +323,92 @@ private CONModification GetCONMod(string name)
             return mods;
         }
 
+
+        private void PreScanUpdateMidis()
+        {
+            YargLogger.MinimumLogLevel = YARG.Core.Logging.LogLevel.Trace;
```
polson: Remove this?

YARG.Core/Song/Entries/MoggAudioLoader.cs
```diff
@@ -0,0 +1,161 @@
...
+{
+    /// <summary>
+    /// Builds a StemMixer directly from a raw, multi-channel .mogg stream plus a
+    /// channel-index/panning map. Originally lived inline in RBCONEntry.LoadAudio;
+    /// factored out so ini/sng entries that ship a .mogg (instead of split per-stem
+    /// files) can use the exact same path.
+    /// </summary>
+    internal static class MoggAudioLoader
+    {
+        public static bool IsSupportedVersion(int version) => version is RBCONEntry.UNENCRYPTED_MOGG or RBCONEntry.YARG_MOGG;
```
polson, line 21: YARG_MOGG is not visible so this doesn't compile

YARG.Core/Song/Entries/MoggAudioLoader.cs
```diff
@@ -0,0 +1,161 @@
...
+            return mixer;
+        }
+
+        /// <summary>
+        /// Parses a single-song .dta fragment (just the `song` block - tracks/pan/vol)
+        /// into channel indices and computed L/R pan gains, using the same math RBCON
+        /// uses. Intended for a small sidecar file that rides next to an ini song's
+        /// raw .mogg, e.g. "song.mogg.dta".
+        /// </summary>
+        public static bool TryParseChannelMap(FixedArray<byte> dtaBytes, out RBAudio<int> indices, out RBAudio<float> panning)
```
polson: This is unused?

YARG.Core/Song/Cache/CacheGroups/IniEntryGroup.cs
```diff
@@ -9,6 +9,7 @@ internal sealed class IniEntryGroup : IEntryGroup
         private readonly string _directory;
         private readonly List<UnpackedIniEntry> _unpacked = new();
         private readonly List<SngEntry> _packed = new();
+        private readonly Dictionary<string, UnpackedIniEntry> _byShortname = new();
```
polson: can two different songs have the same shortname?

polson (changes requested):

asimard1:
When joined with #422, song folders that have a dta file don't need to have a .ini file with a shortname entry in it, since that info can be read directly from the dta node name.

This simplifies the amount of effort required by users to either find an extractor that adds shortnames to the .ini files, or adding them themselves (things I have had to do myself).

PR #413, opened 2026-07-11
Support lanes on Medium, Easy, and Beginner

Frickitickitavi:
[Corresponding YARG PR](https://github.com/YARC-Official/YARG/pull/1557) (details there)

### DIFF
--- a/YARG.Core/Engine/Drums/DrumsEngine.cs
+++ b/YARG.Core/Engine/Drums/DrumsEngine.cs
@@ -28,6 +28,14 @@ namespace YARG.Core.Engine.Drums
 
         protected bool IsMidiDrumsInput;
 
+        protected override int WildcardMask => EngineParameters.Mode switch
+        {
+            DrumsEngineParameters.DrumMode.NonProFourLane or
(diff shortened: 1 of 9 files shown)


YARG.Core/Engine/Drums/DrumsEngine.cs
```diff
@@ -28,6 +28,14 @@ public abstract class DrumsEngine : BaseEngine<DrumNote, DrumsEngineParameters,
...
         protected bool IsMidiDrumsInput;
 
+        protected override int WildcardMask => EngineParameters.Mode switch
+        {
+            DrumsEngineParameters.DrumMode.NonProFourLane or
+            DrumsEngineParameters.DrumMode.ProFourLane => (int) FourLaneDrumPad.Wildcard,
+            DrumsEngineParameters.DrumMode.FiveLane => (int) FiveLaneDrumPad.Wildcard,
+            _ => -1
+        };
+
```
wyrdough: Feels a bit nitpicky, but we could save some cycles by computing this once in the constructor instead of recalculating it every time WildcardMask is called.

wyrdough (changes requested):

PR #414, opened 2026-07-12
Rework EngineContainer to use generics

Supersonicboss1:
We also store an `InstrumentDifficulty<TNoteType>` now, instead of separate Instrument and Difficulty members (though we still expose these through an auto-property for convenience). This has the benefit of `GetUnisonSections` using the note chart the engine uses, instead of the original note chart, preserving any modifications made by modifiers.

We pass in all three parameters the engine requires (note, params, stats) instead of a note/engine pair because it allows us to use anything in `BaseEngine.Generic.cs`, including a bunch of events (e.g. `OnNoteHit`) that we would otherwise not be able to use without manual type checks. This saves us a bunch of boilerplate in event subscriptions, for example (though we still need some, e.g. because of `OnOverhit`/`OnOverstrum`, and vocals-specific events).

Unfortunately a fair bit of boilerplate had to be added to ReplayAnalyzer since it can't really use generics easily.

Requires YARC-Official/YARG#1560.

### DIFF
--- a/YARG.Core.UnitTests/Engine/EngineManagerTester.cs
+++ b/YARG.Core.UnitTests/Engine/EngineManagerTester.cs
@@ -44,10 +44,11 @@ public class EngineManagerTester : EngineTester
     [Test]
     public void EngineManagerGeneratesUnisonPhrases()
     {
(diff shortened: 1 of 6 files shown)


wyrdough (approved):
Gonna need to look at this in context and do a bit of testing, but at first blush this looks good to me. Thanks!

polson (approved):

PR #415, opened 2026-07-13
Add Autodowncharter tool and library calls

jstnlef:
This is essentially a direct port of Hubbubble's python version of the tool. I suspect we'll want to put a good deal more work into tweaking the algorithm but it should be pretty faithful to the original implementation while making use of the game's actual midi parsing. I also wrote it such that the library calls will be easily accessible to the game when we end up implementing that work.

### DIFF
--- a/AutoDownchartCli/AutoDownchartCli.csproj
+++ b/AutoDownchartCli/AutoDownchartCli.csproj
@@ -1,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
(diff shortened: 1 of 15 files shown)


wyrdough (approved):

wyrdough:
Assuming hub is happy enough with the results, I think this is good enough that it can be merged so we can get started with the front end part and get player feedback.

jstnlef:
I believe this should be good to merge as is. Hubbubble does not want this included in YARG proper but the downcharting CLI tool could be useful as a speedup for charters.

PR #419, opened 2026-07-13
Vocals and Drums scoring changes

Supersonicboss1:
This reworks some generally unwanted scoring behavior on Drums and Vocals. If anyone has any issues with the values I chose, or the implementation, please let me know!

## Drums

Ghosts were worth *way* too much score, with both them and accents being a 1.5x additional multiplier on top of the note's score and multiplier. Instead, this PR makes them worth separate amounts. Accents are now a base 25 points, which is roughly 1.5x the note's score and multiplier (which is similar to the old value), but ghost notes are now only a flat 10 extra base points, so at 4x, they score an additional 40 points. This is enough to make a minor difference to scoring, and to differentiate scores on the leaderboards, but not enough to significantly affect stars.

~~Right now, the base score of a note on drums is 50 if not on Pro Drums, and 60 if you are. While this works right now, if we ever allow partial Pro Drums similar to RB, this will cause leaderboard issues. This PR scores 4L and 5L at 50, Elite unconditionally at 60, and Pro Drums at 60 if the note is a cymbal, and 50 otherwise.~~

This also separates DynamicsBonus into separate AccentsBonus and GhostsBonus stats.

## Vocals

Percussion notes were worth far too much relative to the score of the chart, especially on lower difficulties where vocal phrases are worth less. This PR makes it so that, instead of passing in how much a vocal phrase should be worth in EngineParameters, you pass in a score multiplier, which is then applied to both the vocal phrase (2000 * mult) and percussion (100 * mult), scaling it better based on difficulty. Also, this stops percussion notes from being affected by multiplier (which didn't make any sense).

Right now, Star Power on Vocals just doubles the points awarded for a phrase if SP was active when the phrase ended. This is bad because it encourages activating right before the end of a phrase to immediately double your points, and means long vocal phrases can lead you to completely wasting your SP. This PR changes it to work more like RB. It keeps track of the ticks you hit while SP is active, and uses this to determine how many extra points you should be awarded. This also applies to Band Bonus score. It only keeps track until the phrase meter is full, and then any bonus points are ignored.

I also fixed some of the annoying behavior of VocalsStats, namely that TicksHit and TicksMissed do not actually correspond to the absolute number of ticks hit and missed. Instead, if you want the number of phrases hit (aka the effective percent), you can use the PhrasesHit stat which has been added.

I also added some stats for percussion notes, such as the total number of percussion notes, and the number of hit percussion notes.


Ape:
What's the reasoning for ghosts to be worth so much less?

Supersonicboss1:
> What's the reasoning for ghosts to be worth so much less?

Generally speaking charters use a lot more ghosts than they do accents. This was causing issues with allowing the star thresholds to be reached more easily than intended, in charts with a lot of ghosts. FWIW, I was likely a bit overzealous with having them be worth 2 points, but I do think they still need to be worth less.

Ape:
> This was causing issues with allowing the star thresholds to be reached more easily than intended.

If that's the issue, then I think the proper solution is not raise the star thresholds by taking all the potential dynamics bonuses into account.

Supersonicboss1:
> > This was causing issues with allowing the star thresholds to be reached more easily than intended.
> 
> If that's the issue, then I think the proper solution is not raise the star thresholds by taking all the potential dynamics bonuses into account.

The reason I don't want to do that is because I don't want to punish people who have kits which don't support velocity (e.g. I think the RB1 kit doesn't). I know the no dynamics modifier exists, but this would force them onto a separate leaderboard in order to have fair stars which feels wrong.

Also, sorry if it seems like I'm nitpicking. I agree with you that my solution is not ideal, I'm just not convinced there is a better way that does not have some significant tradeoff.

Ape:
> I don't want to punish people who have kits which don't support velocity

But even in your solution they are losing on potential points anyway so the scores are not really comparable or competitive between them anyway. I think we just have to accept that or have separate leaderboard.

Maxio68:
> ...and Pro Drums at 60 if the note is a cymbal, and 50 otherwise.

This is not how Rock Band handles cymbals scoring - looking at videos, *every* note scores 30 with all three cymbals; and as far as I remember from looking online, notes would score 25 if they *were* cymbals, but you didn't have that cymbal enabled. [Ghostbusters opens with a roll on the Blue tom,](https://www.youtube.com/watch?v=vBv4t-sifB0) which all score 30 points. (These values are doubled in YARG to line up with Clone Hero scoring, and all notes were only worth 25 when outside of Pro Drums.)

> ...but ghost notes are now only a flat 2 extra base points

I'd probably put them at 10 extra base points.

Supersonicboss1:
> > I don't want to punish people who have kits which don't support velocity
> 
> But even in your solution they are losing on potential points anyway so the scores are not really comparable or competitive between them anyway. I think we just have to accept that or have separate leaderboard.

Fair. My goal here (although obviously you or the YARG team may not agree) is that dynamics can act as a differentiator between two similarly good scores (e.g. two similar FCs) but generally they don't allow a worse score if dynamics were ignored to place above a significantly better one. 

There is definitely an argument to be made that this is not the correct direction, but in my experience drum dynamics are a bit of a inconsistent feature, both in terms of how well they work on a given kit, and how a charter uses them, so I don't really want them to be incredibly significant (and in terms of not factoring them into star thresholds, I would like to lean towards being more generous to the user, since dynamics aren't a feature you are ever required to interact with). 

FWIW I think Elite Drums is likely to place more stock in dynamics, since ghost notes work differently in it, and it has been designed from the beginning with the understanding they would be a more important feature.


> This is not how Rock Band handles cymbals scoring - looking at videos, _every_ note scores 30 with all three cymbals; and as far as I remember from looking online, notes would score 25 if they _were_ cymbals, but you didn't have that cymbal enabled. [Ghostbusters opens with a roll on the Blue tom,](https://www.youtube.com/watch?v=vBv4t-sifB0) which all score 30 points. (These values are doubled in YARG to line up with Clone Hero scoring, and all notes were only worth 25 when outside of Pro Drums.)

Really? I had no idea, I was under the impression that only the cymbals score extra. Thanks for the tip, will fix (which ironically makes it the same as before)

> > ...but ghost notes are now only a flat 2 extra base points
> 
> I'd probably put them at 10 extra base points.

Will do, this seems like a reasonable number (though obviously the other discussion is not settled yet, so we'll see where that leads).

PR #420, opened 2026-07-16
rid of warnings

theli-ua:


YARG.Core/Game/YargProfile.cs
```diff
@@ -474,22 +474,31 @@ public void Serialize(BinaryWriter writer)
 
             writer.Write((byte) OpenLaneDisplayType);
 
-            writer.Write((byte)FourLaneDrumsHighwayOrdering.Length);
-            foreach (var item in FourLaneDrumsHighwayOrdering)
+            if (FourLaneDrumsHighwayOrdering != null)
             {
-                writer.Write((byte) item);
+                writer.Write((byte) FourLaneDrumsHighwayOrdering!.Length);
```
polson: I am pretty sure skipping writes if something is null is going to mess stuff up.  I think we need to write a default value here if something is null?

polson (changes requested):

YARG.Core/Game/YargProfile.cs
theli-ua: You're right of course, its a brainfart. 
It would break if this would actually happen as deserialization reads them unconditionally.

My bad for blindly just being annoyed by warnings without putting any thought into it.

PR #422, opened 2026-07-18
Add mogg, dta and png_xbox support to ini formats

asimard1:
Add checks to see if mogg and png_xbox files are present in the directory of a song.

Also refactored the mogg mixer logic to a new file so that both RBCONs and ini could use it.


asimard1:
I want to add dta support as well, so that every single file that is in the CON can be read directly if extracted as is.

asimard1:
Is it worth adding dta support as well? That way all files can be read directly as extracted without modification.

wyrdough:
Why? I fail to see the point of this when exCONs are already supported.

asimard1:
The main reason is that supporting the original `.mogg` and `.png_xbox` files makes extracting CON packs more practical without requiring users to convert any files. I have found this to be useful because:

- ~~CON multipacks don't have video background support in YARG, so extracting the songs is currently the only way to get video backgrounds working. This requires tools like Onyx or others, but these convert files and it's nice to have the option to use the files directly as-is.~~ (This was wrong.)
- CON multipacks aren't categorized properly when sorting by folders. Extracting them solves this problem. See #454.
- MOGG extractors don't always produce an ideal result. The resulting audio can either lose quality or be significantly larger than the original MOGG.
- Also, in my experience, some extractions using Onyx have had audio syncing problems between the stems it produced. This is not a problem that the YARG team should fix, but it's nice to have other options.
- DTA parsing isn't always perfect either when creating ini files. I also added direct `.dta` support, which removes the need for external DTA parsing.
- More generally, supporting the original file formats reduces the number of external conversion/extraction steps required and lets YARG work with the data as it actually exists in CON packs.
- The final point is that converting CON packs to stems takes longer than just extracting the mogg file, so this feature is nice to have for huge libraries.

Obviously, exCONs are still a valuable option to have for most people, but there are cases where what I'm adding is useful, and I'm sure other people would prefer to use the original files without converting them first.

Let me know if you disagree, my experience using this PR has been very positive personally.

asimard1:
The only change missing for this is to add dta detection during scan. At the moment, if a song is missing a .ini file, it won't get seen at all, but if it has, the .dta is completely ignored. I will do that tomorrow probably.

asimard1:
I found while working on #454 that CON multisong packs actually do support video backgrounds when the videos are placed alongside the CON and named using the `$"{dtanode}.{extension}"` convention. The folder-sorting issue is also being addressed in that PR.

This means two of the reasons I originally mentioned here as motivation are less significant now. I still think direct `mogg` and `png_xbox` support is worthwhile for the other reasons discussed in my previous comment, so I will leave #422 open.

I still believe that supporting the maximum number of file types will help make the game more and more accessible, but I understand that this also means more code to maintain. #454 should be considerably less intrusive.

PR #425, opened 2026-07-20
Implement per-stem drum muting

DevPrice:
This PR adds support in core for muting individual drum stems based on which gems are missed.

See the YARC-Official/YARG#1573 for more information about the overall feature.

The changes for Core are:
* Split `SongStem.Drums` into `Drums1`, `Drums2`, and `Drums3`
  * This approach was chosen over using more semantic names (e.g., `DrumsKick`, `DrumsSnare`, etc.) because the architecture today assumes that filenames each map to a unique `SongStem`, but drum stems can vary based on mix configuration events in the chart. It also makes it easier to have fallback logic for when the stem files don't exactly match the expected mix configuration (e.g., the chart says it supports 3 stems, but the player actually only has a single combined stem).
* Add a `DrumStem` enum, and store and expose the associated `DrumStem` for each `DrumNote`.
* Update drums `MoonSongLoader` to associate a drum stem with each drum note based on its mix configuration, disco flip state, and gem.
* Replace `Instrument.ToSongStem()` with `Instrument.ToSongStems()`, allowing an instrument to be associated with more than one stem


YARG.Core/Audio/AudioHelpers.cs
```diff
@@ -43,45 +43,52 @@ public static SongStem GetStemFromName(string stem)
...
             };
         }
 
-        public static SongStem ToSongStem(this Instrument instrument)
+        private static readonly ISet<SongStem> GuitarStems  = new HashSet<SongStem> { SongStem.Guitar };
+        private static readonly ISet<SongStem> BassStems    = new HashSet<SongStem> { SongStem.Bass };
+        private static readonly ISet<SongStem> RhythmStems  = new HashSet<SongStem> { SongStem.Rhythm };
+        private static readonly ISet<SongStem> KeysStems    = new HashSet<SongStem> { SongStem.Keys };
+        private static readonly ISet<SongStem> DrumsStems   = new HashSet<SongStem> { SongStem.Drums1, SongStem.Drums2, SongStem.Drums3 };
+        private static readonly ISet<SongStem> VocalsStems  = new HashSet<SongStem> { SongStem.Vocals };
```
wyrdough: Is there a reason to pay the indirection penalty of declaring these as `ISet` instead of straight `HashSet`?

polson: I assume they did this to hide the mutability from callers

DevPrice: That was my intent, but then I realized that `ISet` still exposes all the mutation methods

PR #435, opened 2026-07-29
EngineManager unregistration changes

Supersonicboss1:
These changes are in service of the goals to
1. Make EngineManager more capable and reliable when removing engines mid-gameplay
2. Fixing state mismanagement when unregistering and re-registering engines with the same EngineManager instance (practice mode)

To this effect, we now actually unsubscribe from engine events to avoid a memory leak when engines are recreated, unison events are now removed entirely if they have no participants from an unregistration, and there are now some fixes regarding coda and fail meter state when resetting EngineManager state (replay or practice).

Also, `_allEnginesById` is now removed. I don't think it was very useful, and was just extra state to keep track of.

Required by YARC-Official/YARG#1595


wyrdough (approved):
lgtm, so yolo

PR #439, opened 2026-08-02
Add tests, fix scoring bugs

polson:
Note: This pr was almost entirely vibecoded as I just wanted to get some tests in for the scoring, given we have more than one PR to do with scoring changes.  In the process, two bugs may have been found.  I appreciate any look at the scoring changes as I'm not entirely sure what is correct there.

 Added tests for:
 - DrumsEngine
 - GuitarEngine
 - FiveLaneKeysEngine
 - ReplayInfo
 - MidiPreparser

 Bug 1 (5-lane keys): the song's max score counted every sustain note's points twice.
 Introduced in https://github.com/YARC-Official/YARG.Core/commit/b3d37169

 Bug 2 (guitar): split chords got extra combo in the max-score math, but the game only gives one combo per strum.
 Introduced in  https://github.com/YARC-Official/YARG.Core/commit/df171b81

### DIFF
--- a/YARG.Core.UnitTests/Engine/DrumEngineTester.cs
+++ b/YARG.Core.UnitTests/Engine/DrumEngineTester.cs
@@ -54,6 +54,174 @@ public class DrumEngineTester : EngineTester
     }
 
     [Test]
+    public void TenSingleNotes_FcUsesCorrectMultiplierBoundary()
+    {
+        var notes = Enumerable.Range(1, 10)
+            .Select(index => new DrumNote(
+                FourLaneDrumPad.RedDrum,
+                DrumNoteType.Neutral,
+                DrumNoteFlags.None,
+                NoteFlags.None,
+                index * 0.5,
+                (uint) index * 480))
+            .ToList();
+        LinkNotes(notes);
+
+        var difficulty = new InstrumentDifficulty<DrumNote>(Instrument.ProDrums, Difficulty.Expert,
+            notes, new(), new());
+        var syncTrack = new SyncTrack(480);
+        syncTrack.Tempos.Add(new TempoChange(120, 0, 0));
+        var engineParams = new DrumsEngineParameters(
+            new HitWindowSettings(0.1, 0.1, 1.0, false, 0, 1.0, 1.0, 0.15, 0.25),
+            4,
+            StarMultiplierThresholds,
+            SoloBonusStarMultiplierThresholds,
+            DrumsEngineParameters.DrumMode.ProFourLane,
+            false,
+            true);
+        var engine = new YargDrumsEngine(difficulty, syncTrack, engineParams, true, false);
+
+        engine.Update(notes[^1].Time + 0.5);
+
+        using (Assert.EnterMultipleScope())
+        {
+            // 10 notes at 60 points each (pro four lane = 50 + 10); chart base score assumes
+            // the pre-note combo (all 1x), while an FC commits the 10th note at 2x
+            Assert.That(engine.BaseNoteScore, Is.EqualTo(600));
+            Assert.That(engine.BaseScore, Is.EqualTo(600));
+            Assert.That(engine.EngineStats.CommittedScore, Is.EqualTo(660));
+            Assert.That(engine.EngineStats.NoteScore, Is.EqualTo(600));
+            Assert.That(engine.EngineStats.MultiplierScore, Is.EqualTo(60));
+            Assert.That(engine.EngineStats.PendingScore, Is.Zero);
+            Assert.That(engine.EngineStats.TotalNotes, Is.EqualTo(10));
+            Assert.That(engine.EngineStats.NotesHit, Is.EqualTo(10));
+            Assert.That(engine.EngineStats.NotesMissed, Is.Zero);
+            Assert.That(engine.EngineStats.MaxCombo, Is.EqualTo(10));
+            Assert.That(engine.EngineStats.Percent, Is.EqualTo(1f));
+            Assert.That(engine.EngineStats.IsFullCombo, Is.True);
+        }
+    }
+
+    [Test]
+    public void MissOneNote_PercentReflectsMissAndComboResets()
+    {
+        var notes = Enumerable.Range(1, 10)
+            .Select(index => new DrumNote(
+                FourLaneDrumPad.RedDrum,
+                DrumNoteType.Neutral,
+                DrumNoteFlags.None,
+                NoteFlags.None,
+                index * 0.5,
+                (uint) index * 480))
+            .ToList();
+        LinkNotes(notes);
+
+        var difficulty = new InstrumentDifficulty<DrumNote>(Instrument.ProDrums, Difficulty.Expert,
+            notes, new(), new());
+        var syncTrack = new SyncTrack(480);
+        syncTrack.Tempos.Add(new TempoChange(120, 0, 0));
+        var engine = new YargDrumsEngine(difficulty, syncTrack, _engineParams, false, false);
+
+        // Hit every note except the 5th (at 2.5s)
+        for (int i = 0; i < notes.Count; i++)
+        {
+            if (i == 4)
+            {
+                continue;
+            }
+
+            HitPad(engine, notes[i].Time, DrumsAction.RedDrum);
+        }
+
+        engine.Update(notes[^1].Time + 0.5);
+
+        using (Assert.EnterMultipleScope())
+        {
+            Assert.That(engine.EngineStats.NotesHit, Is.EqualTo(9));
+            Assert.That(engine.EngineStats.NotesMissed, Is.EqualTo(1));
+            Assert.That(engine.EngineStats.MaxCombo, Is.EqualTo(5));
+            Assert.That(engine.EngineStats.Percent, Is.EqualTo(0.9f));
+            Assert.That(engine.EngineStats.IsFullCombo, Is.False);
+        }
+    }
+
+    [Test]
+    public void SoloUnderSixtyPercent_AwardsNoBonus()
+    {
+        // 5/10 = 50% < 60% -> bonus deleted
+        var (engine, notes) = CreateSoloEngine(10);
+
+        for (int i = 0; i < 5; i++)
+        {
+            HitPad(engine, notes.Notes[i].Time, DrumsAction.RedDrum);
+        }
+
+        engine.Update(notes.Notes[^1].Time + 0.5);
+
+        Assert.That(engine.EngineStats.SoloBonuses, Is.Zero);
+    }
+
+    [Test]
+    public void SoloAtExactlySixtyPercent_AwardsNoBonus()
+    {
+        // 6/10 = 60% -> multiplier clamps to 0 -> 0 points
+        var (engine, notes) = CreateSoloEngine(10);
+
+        for (int i = 0; i < 6; i++)
+        {
+            HitPad(engine, notes.Notes[i].Time, DrumsAction.RedDrum);
+        }
+
+        engine.Update(notes.Notes[^1].Time + 0.5);
+
+        Assert.That(engine.EngineStats.SoloBonuses, Is.Zero);
+    }
+
+    [Test]
+    public void SoloOverSixtyPercent_RoundsBonusDownToNearestFifty()
+    {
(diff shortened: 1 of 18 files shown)


YARG.Core/Engine/Keys/FiveLaneKeys/FiveLaneKeysEngine.cs
```diff
@@ -181,38 +181,16 @@ protected sealed override (int baseScore, int noteScore) CalculateChartScores()
...
-                        {
-                            combo++;
-                            seenNoteTicks.Add(child.Tick);
-                        }
-                    }
+                    multiplier = Math.Min((combo / 10) + 1, BaseParameters.MaxMultiplier);
+                    double scoreForNote = POINTS_PER_NOTE + (int) Math.Ceiling(chordNote.TickLength / TicksPerSustainPoint);
+                    baseScore += multiplier * scoreForNote;
+                    noteScore += scoreForNote;
+                    combo++;
```
Supersonicboss1: Somewhat counterintuitively, combo and multiplier are not linked on Keys (5LK and PK) the same way they are on other instruments. Even though a chord counts each note separately for combo, this does not apply to the multiplier, which counts an entire chord as one note, so the multiplier should not increase with each note in the chord here.

YARG.Core.UnitTests/Engine/DrumEngineTester.cs
```diff
@@ -332,6 +500,22 @@ private static (YargDrumsEngine Engine, InstrumentDifficulty<DrumNote> Notes) Cr
     }
 
 
+    private static void LinkNotes(List<DrumNote> notes)
```
Supersonicboss1: Bit of a nitpick, but this function can probably prove useful for other engine testers, since they all use the same note structure (other than vocals, to some extent). I would probably extract it as a generic function e.g. `private static void LinkNotes<TNote>(List<TNote> notes) where TNote : Note<TNote>`

YARG.Core.UnitTests/Replays/ReplayInfoTests.cs
```diff
@@ -0,0 +1,255 @@
...
+        typeof(ReplayInfo).GetField(nameof(ReplayInfo.ReplayName))!.SetValue(info,
+            ReplayInfo.ConstructReplayName(info.SongName, info.ArtistName, info.CharterName, info.Date));
+        return info;
+    }
+
+    private static object DistinctValueFor(Type type, string fieldName)
+    {
+        if (type == typeof(string))
+        {
+            return "pǝɹʇ-юникод-名前";
```
Supersonicboss1: If this is intentional (e.g. to test weird values), you should probably add a comment informing that. If not, then???

Supersonicboss1 (changes requested):

YARG.Core/Engine/Keys/FiveLaneKeys/FiveLaneKeysEngine.cs
polson: Thanks for your review.  This PR is heavily vibecoded and I definitely need eyes on the scoring changes here before I do anything else with it

YARG.Core.UnitTests/Replays/ReplayInfoTests.cs
polson: yeah this is a bit extra... I'll change it

YARG.Core/Engine/Keys/FiveLaneKeys/FiveLaneKeysEngine.cs
polson: @Supersonicboss1 okay I'm trying to understand what should be correct here.

Lets start with a chord, if I hit a 2 chord should that be +2 combo or +1 combo?

Supersonicboss1: For this function, +1. Basically on keys, for the purpose of the multiplier only, each chord is +1 combo.

YARG.Core/Engine/Guitar/GuitarEngine.cs
Supersonicboss1, line 423: This should be a `LogFormatDebug` for performance reasons, though this is an issue on all engines with this override, so if you don't want to do it in this PR, that's fine

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -421,20 +420,15 @@ protected sealed override (int, int) CalculateChartScores()
...
+                // Disjoint chords are hit with a single strum, so they do not
+                // increment combo beyond the single increment above.
                 if (note.IsDisjoint)
                 {
-                    HashSet<uint> seenNoteTicks = new();
                     foreach (var child in note.ChildNotes)
                     {
                         double pointsForDisjoint = Math.Ceiling(child.TickLength / TicksPerSustainPoint);
                         baseScore += multiplier * pointsForDisjoint;
                         noteScore += pointsForDisjoint;
```
Supersonicboss1: This is not your fault, but the sustain points calculation has been broken for a while, since we don't count child notes unless it's a disjoint (extended) chord. This code probably needs to lose the `note.IsDisjoint` check and iterate over `note.AllNotes`, and remove the extra sustain point calculations above.

YARG.Core/Engine/Guitar/GuitarEngine.cs
```diff
@@ -421,20 +420,15 @@ protected sealed override (int, int) CalculateChartScores()
                 noteScore += pointsForSustain;
                 combo++;
```
Supersonicboss1, line 410: We need to recalculate the multiplier again here, since it may be different for the sustains vs. the notes themselves since the combo has increased.

Supersonicboss1 (changes requested):
Some more changes I missed in my first review. For how the score threshold calculation code works, for all of (Guitar, 5LK, Pro Keys) the Pro Keys engine has a (mostly) correct implementation you can follow (though it also has the issue I mentioned in the guitar engine where we don't recalculate multiplier after increasing combo, and it increases combo after sustain calculations which is wrong)

YARG.Core/Engine/Guitar/GuitarEngine.cs
polson: Ok so, right now I'm trying to learn scoring because it seems there is no clear source of truth for what scoring should do other than the code itself, which is a problem.  From what I can tell, there is no case where hitting a chord increases combo by more than one.

Edit: I guess for drums, a "chord" is going to increase combo for all notes hit: `1 + ChildNotes.Count`

As for sustain points, no idea, I'll try to figure out exactly how that should work and if we can fix something here.

YARG.Core/Engine/Guitar/GuitarEngine.cs
polson: Appreciate you looking at this as always.  Let me actually learn how scoring works before I do anything, otherwise I'm just slopcoding blindly

Supersonicboss1:
@polson To my knowledge, this is how scoring works (excluding Vocals).

Hitting a note increments your combo by 1. Your multiplier increases every 10 combo, on the 10th note, for a max of 4 (or 6, on Bass). Chords (multiple notes on the same tick) count as 1 combo on all instruments **except Drums,** where each note is counted separately.
e.g.
0-9 => 1x
10-19 => 2x
20-29 => 3x
30-39 => 4x

Combo, and therefore multiplier, is incremented ~~after~~ **before** the entire note is hit, so if a note hit increases the multiplier from 2x to 3x, for example, it will be scored at ~~2x~~ 3x. Sustains are scored based on their tick length (25 points per beat, multiplied by current multiplier).

Disjoint chords are chord sustains where each note's sustain in the chord are not equal in length (e.g. an extended sustain).

The score a note gives should be (except for on drums): (base score per note * number of notes in the chord * multiplier) + (total sustain ticks across all notes in chord * multiplier / `TicksPerSustainPoint`)

I'm not sure exactly how the multiplier increment is handled on drums. I think this method on the drums engine is not exactly correct, since it "hits" all the notes, then increments the combo all at once, which would be wrong in the case that the multiplier increases partway through a chord.

FWIW this method is only used to inform the star thresholds, so it being wrong is not actually that important, since it doesn't really affect gameplay. This is why it (intentionally) does not take into account Star Power.

Note: on drums, notes on the same tick being counted as a "chord" is purely an implementation detail and convenience feature, it means nothing to the way the drums engine works, which treats every note as strictly separate and not affected by the others (except in the case of the dynamic hit window, I guess).

YARG.Core/Engine/Guitar/GuitarEngine.cs
polson: @Supersonicboss1 ok sounds like for sustains you are saying:

1. Sustain points are awarded per note, sustain can be disjoint, and sustain can be a chord
2. Each sustained note in a chord has full-hold points: `ceil(note.TickLength / TicksPerSustainPoint)` 
3. Chord sustain total then is the sum of sustain note points

This code is currently only applying this logic when the chord is disjoint, otherwise it is only awarding points for a single note's sustain in a chord

Is that right?

polson:
> The score a note gives should be (except for on drums): (base score per note * number of notes in the chord * multiplier before hit) + (total sustain ticks across all notes in chord * multiplier after hit / TicksPerSustainPoint)

@Supersonicboss1 maybe I am not understanding something.  This does not seem to be the case for normal scoring.  Normally combo note 10 is worth 2x, here you are saying it would be worth 1x.  Why does it diverge for star cutoffs?  Why not calculate the same way?

Supersonicboss1:
> > The score a note gives should be (except for on drums): (base score per note * number of notes in the chord * multiplier before hit) + (total sustain ticks across all notes in chord * multiplier after hit / TicksPerSustainPoint)
> 
> @Supersonicboss1 maybe I am not understanding something. This does not seem to be the case for normal scoring. Normally combo note 10 is worth 2x, here you are saying it would be worth 1x. Why does it diverge for star cutoffs? Why not calculate the same way?

Sorry, ignore me. You're right, I should have checked the code more closely - the combo should increment before any score calculations are done, at least on Guitar and Drums. Though, this may not be the case for Pro Keys and 5LK, since each note is hit seperately, but combo only increments once the chord is fully hit (I have not checked).

polson:
> Sorry, ignore me. You're right, I should have checked the code more closely - the combo should increment before any score calculations are done, at least on Guitar and Drums. Though, this may not be the case for Pro Keys and 5LK, since each note is hit seperately, but combo only increments once the chord is fully hit (I have not checked).

No worries at all, I'm trying to decipher all this as I go.  What I am seeing is that for star cutoffs, we do a simulation which is a bit different than the normal scoring.  I'm wondering now, if it would be worth it to just do a real simulation with our engine, and then we don't have two different scoring mechanisms that diverge.

Supersonicboss1:
> > Sorry, ignore me. You're right, I should have checked the code more closely - the combo should increment before any score calculations are done, at least on Guitar and Drums. Though, this may not be the case for Pro Keys and 5LK, since each note is hit seperately, but combo only increments once the chord is fully hit (I have not checked).
> 
> 
> 
> No worries at all, I'm trying to decipher all this as I go.  What I am seeing is that for star cutoffs, we do a simulation which is a bit different than the normal scoring.  I'm wondering now, if it would be worth it to just do a real simulation with our engine, and then we don't have two different scoring mechanisms that diverge.

I considered that, but I don't think it would be worth the performance/loading time penalty, especially for what is relatively unimportant consistency-wise. I guess there might be a way to do a faster simulation but hook more into the engine loop, but again, idk if it's worth it.

04meta:
I recently independently found out about the 5-lane keys bugs (both the sustain thing and the fact that every note counts twice towards multiplier, both of which this PR already fixes), as this turns out to make gold stars impossible on anything with a lot of sustains; it's also at least a little funny seeing a bot get 4 stars. I was about to file an issue about it, but with this PR already here, there's no point.

> Though, [the combo maybe shouldn't increment before scoring] for Pro Keys and 5LK, since each note is hit seperately, but combo only increments once the chord is fully hit (I have not checked).

I just checked, and it appears that the engine (both in 5-lane and Pro) waits to see if you hit the entire chord before scoring anything, and scores the whole chord at the new multiplier if applicable:

https://github.com/user-attachments/assets/d5f54741-d839-4787-a0c2-1fc296c86ca0

This means that, at least for an FC with no star power (conceptually the base score), the score matches guitar even if you don't hit all the notes at the same time, so there's no issue here... I think?

PR #440, opened 2026-08-07
Support for multichannel mic handling

polson:
Support for https://github.com/YARC-Official/YARG/pull/1610/changes

### DIFF
--- a/YARG.Core/Audio/GlobalAudioHandler.cs
+++ b/YARG.Core/Audio/GlobalAudioHandler.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 using YARG.Core.Logging;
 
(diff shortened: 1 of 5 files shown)


YARG.Core/Audio/GlobalAudioHandler.cs
```diff
@@ -496,27 +497,29 @@ public static int CreateMetronomeStream(MetronomeSample sample, MetronomePitch p
             }
         }
 
-        public static List<(int id, string name)> GetAllInputDevices()
+        public static Task<List<InputDeviceInfo>> GetAllInputDevicesAsync(CancellationToken cancellationToken = default)
```
wyrdough: Use `UniTask`, not C# `Task`. It generates less garbage (assuming we have already run at least one task that has since completed).

wyrdough: Oh wait, this is core, n/m (I mean, it would still be better to use UniTask, but we don't want to make it a dependency in core)

polson: I could move this logic into YARG, it's just a little messier, but we could UniTask it then

PR #443, opened 2026-08-12
Read CON listings in parallel during cache deserialization

DevPrice:
GetCacheCONListings held the cacheCONListings lock across the file open and the listing parse itself. Since every caller reaches it from a Parallel.ForEach over the CON groups in the cache, that serialized the entire library behind one file read at a time where every thread waited on the lock.

Only the dictionary lookup needs to be guarded. Storing a `Lazy<>` lets the read happen outside the lock while still parsing any individual CON exactly once.

Also gives the `FileStream` a buffer large enough to cover the CON header. The header is read in four small chunks spread over the first 0x380 bytes, and with the previous buffer size of 1 each of those was its own trip to the filesystem.

Measured on a 4187 song library (3394 ini, 175 sng, 618 packed CONs) where the CONs live on a network share:

  cold cache:  6362ms -> 812ms
  warm cache:   186ms ->  61ms

The win scales with how many separate CON files a library has, and is largest when they are slow to reach.

### DIFF
--- a/YARG.Core/Song/Cache/CacheHandler.cs
+++ b/YARG.Core/Song/Cache/CacheHandler.cs
@@ -905,7 +905,13 @@ namespace YARG.Core.Song.Cache
         private HashSet<string> invalidSongsInCache = new();
         private Dictionary<string, FileCollection> collectionCache = new();
         private Dictionary<string, QuickCONMods> cacheCONModifications = new();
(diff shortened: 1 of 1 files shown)


wyrdough:
Thanks, this appears to be a nice speedup!

PR #449, opened 2026-08-31
Improved auto-generated lipsync

theli-ua:
### Improved auto-generated lipsync 
 Visemes
 - Corrected the phoneme ->  viseme mapping
 - Long sustained notes keep the mouth open for the whole note 
 - Mouth openness varies slightly with pitch while sustaining
 - Transitions between mouth shapes flow with a brief overlap instead of snapping
 - The mouth closes fully between syllables and during instrumental breaks
 
 Facial expression
 - Replaced random per-phrase expression blips with sustained emotional states (master version is too erratic)
 - Broader expression palette, including squints and brow variation

Here is one example track from the setlist: 
https://youtu.be/y3TgKjafbds

I've also shared another comparison track [on discord](https://discord.com/channels/1086048856678084609/1086052453801283624/1543799936561451178)


theli-ua:
Here's one example where I think the PR version may be better than authored lipsync in the chart in the setlist.

image

The "you" looks somewhat off in authored data

Supersonicboss1:
This isn't a review, since this PR is still a draft, but I should let you know that the static lyric shifts PR that got merged recently assigned lengths to lyric events on any vocals track (though not the lyrics track itself), which may allow you to simplify somewhat here.

polson (approved):

PR #450, opened 2026-09-02
Handle open inputs in BREs

Frickitickitavi:
[Corresponding YARG PR](https://github.com/YARC-Official/YARG/pull/1654) (details there)


Supersonicboss1 (approved):
You're right that this is somewhat hacky, but I think it's fine for what it is.

PR #452, opened 2026-09-04
fix(ultrastar): Properly follow Ultrastar file naming conventions, other fixes

taynaron:
Extends existing UltraStar support (https://github.com/YARC-Official/YARG.Core/pull/343, https://github.com/YARC-Official/YARG.Core/pull/424) in YARG.Core, making it compliant with the actual spec ([Overview](https://github.com/UltraStar-Deluxe/Play/wiki/UltraStar-txt-File-Format), [Details](https://github.com/UltraStar-Deluxe/format/blob/main/The%20UltraStar%20File%20Format%20(v1).md)).

### Feature additions

- Read audio/video/cover/background from #AUDIO/#MP3/#VIDEO/#BACKGROUND/#COVER tags instead of fixed filenames - the biggest change needed to make YARG compliant with UltraStar format
  - folders with multiple .txt charts now scan each as its own song
- Handle #BPM tempo changes mid-song (B lines), not just the initial value
- Support a third voice/harmony part (P3), capped at 3 to match HARM1/2/3
- Map #EDITION to Source (the FoF "icon" field's equivalent slot — actual text/icon display still needs work in the game repo)
- Map #COMMENT to the loading phrase

### Bug fixes

- Comma-decimal tag values (#VIDEOGAP:1,5) now parse correctly
- #GAP was being double-applied (baked into note ticks and mapped to SongOffset), causing lyrics/video to drift out of sync with audio
- #VIDEOGAP sign was inverted
- Freestyle (F)/rap (R)/golden-rap (G) notes now rendered as unpitched 
- Trailing ~ (melisma) didn't structurally merge into the following note
- Stray .txt files (readme/license) in a chart folder no longer count as broken songs
- Blank tags (e.g. #ARTIST:) no longer overwrite defaults
- Case- and Unicode-normalization-insensitive subfile/audio/video resolution (fixes songs with accented filenames on macOS)
- Handled weird cases like a Freestyle `~`
- A chart whose audio tag points at a video file (rare, no separate audio track) now fails the scan with a clear message in badsongs.txt instead of silently entering the library and failing at play time

Don't be afraid of the diff size, a lot of it is regression testing

### DIFF
--- a/YARG.Core/Chart/Loaders/UltraStar/UltraStarLoader.cs
+++ b/YARG.Core/Chart/Loaders/UltraStar/UltraStarLoader.cs
@@ -13,10 +13,15 @@ namespace YARG.Core.Chart.Loaders.UltraStar
     {
         #region Constants
 
-        // MIDI base constant
-        // UltraStar pitch is RELATIVE
-        // MIDI 60 = C4
+        // UltraStar pitch is relative to C4 (MIDI 60).
         private const int ULTRASTAR_PITCH_BASE = 60;
+
+        // Melisma/continuation marker on a syllable. Distinct from the rest marker '-',
+        // which shares a character with LyricSymbols.LYRIC_JOIN_SYMBOL but is unrelated.
+        private const char US_MELISMA_SYMBOL = '~';
+
+        // An UltraStar beat is an eighth of the internal tick beat.
+        private const uint US_BEATS_PER_TICK_BEAT = 8;
 
         #endregion
 
@@ -42,8 +47,9 @@ namespace YARG.Core.Chart.Loaders.UltraStar
         private readonly List<(uint Beat, double Bpm)> _tempoChanges = new();
 
         private readonly Dictionary<int, List<UltraStarNote>> _partNotes = new();
-        private readonly HashSet<int> _seenParts = new();
         private int _currentPart = 0;
+        // Set by a trailing '~'; the next note consumes it as its pitch-slide marker.
+        private bool _pendingPitchSlide = false;
 
         #endregion
 
@@ -51,7 +57,6 @@ namespace YARG.Core.Chart.Loaders.UltraStar
 
         private class UltraStarNote
         {
-            public int    PartIndex     { get; set; } = 0;
             public char   Type          { get; set; }
             public uint   StartBeat     { get; set; }
             public uint   DurationBeats { get; set; }
@@ -59,15 +64,26 @@ namespace YARG.Core.Chart.Loaders.UltraStar
             public string Lyric         { get; set; } = string.Empty;
 
             /// <summary>
-            /// When true, a hyphen ('-') should be appended to this note's lyric
-            /// and JoinWithNext flag set on its LyricEvent. Set when the NEXT note
-            /// in the phrase has a '~' (melisma continuation) prefix.
+            /// Hyphenates this note's lyric onto the next one (JoinWithNext). Set either by
+            /// a trailing '~' here or by a leading '~' with text on the following note.
             /// </summary>
             public bool MelismaJoin { get; set; }
 
-            public bool IsGolden    => Type == '*' || Type == 'G';
-            public bool IsUnpitched => Type == 'F' || Type == 'R' || Type == 'G';
-            public bool IsRest      => Type == '-';
+            public uint EndBeat => StartBeat + DurationBeats;
+
+            // Type rules live here so adding a note type touches one place.
+            public static bool IsNoteLineType(char type) => type is ':' or '*' or 'F' or '-' or 'R' or 'G';
+            public static bool IsGoldenType(char type)   => type is '*' or 'G';
+            public static bool IsRestType(char type)     => type == '-';
+
+            // Freestyle (F), Rap (R), Golden Rap (G) carry no pitch requirement. All three
+            // are treated as scored+unpitched; per spec Freestyle should be unscored, but
+            // YARG has no zero-score vocal category (see VocalNote.IsNonPitched).
+            public static bool IsUnpitchedType(char type) => type is 'F' or 'R' or 'G';
+
+            public bool IsGolden    => IsGoldenType(Type);
+            public bool IsUnpitched => IsUnpitchedType(Type);
+            public bool IsRest      => IsRestType(Type);
         }
 
         #endregion
@@ -79,6 +95,17 @@ namespace YARG.Core.Chart.Loaders.UltraStar
 
         public string? GetMetadata(string key)
             => _metadata.TryGetValue(key, out var v) ? v : null;
+
+        /// <summary>Voices the note body actually uses; not the #PARTS tag's value.</summary>
+        public int VoiceCount => _partNotes.Count;
+
+        /// <summary>Parses a numeric tag. US files routinely use comma decimals.</summary>
+        public static bool TryParseNumber(string? raw, out double value)
+        {
+            value = 0;
+            return raw != null
+                && double.TryParse(raw.Replace(',', '.'), NumberStyles.Float, CultureInfo.InvariantCulture, out value);
+        }
 
         #region Parsing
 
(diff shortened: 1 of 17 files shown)


taynaron:
@Nort1346 you would likely be interested in this, do you have any thoughts?

YARG.Core/Song/Entries/Ini/SongEntry.IniBase.cs
```diff
@@ -567,29 +577,60 @@ private static ScanResult ScanUltraStar(IniSubEntry entry, FixedArray<byte> file
...
+                return !string.IsNullOrWhiteSpace(value) ? value : fallback;
+            }
+
+            // #AUDIO is the canonical tag; #MP3 is the legacy synonym some tooling still
+            // writes (the file isn't necessarily an mp3).
+            string? audioFile = Tag("AUDIO") ?? Tag("MP3");
+            if (audioFile == null || !SubFileExists(entry._location, audioFile))
+            {
+                return ScanResult.NoAudio;
+            }
```
Supersonicboss1, line 593: We don't need to care about this here - enumerating a directory is slow and involves a bunch of filesystem hits, and we check this in CacheHandler anyway.

YARG.Core/Song/Cache/CacheHandler.cs
```diff
@@ -806,51 +810,110 @@ private void ScanFile(FileInfo info, IniEntryGroup group, in PlaylistTracker tra
...
                     }
                 }
-                catch (PathTooLongException)
+            }
+            catch (Exception e)
+            {
+                YargLogger.LogException(e, $"Error while probing text file {file}!");
+            }
+            return false;
+        }
```
Supersonicboss1, line 890: I'm iffy on this. Having to open and read the contents of every txt file in order to find which is a valid chart is likely to be slow. See my other comment for more details.

YARG.Core/Chart/Loaders/UltraStar/UltraStarLoader.cs
```diff
@@ -160,14 +222,13 @@ private void ParseNoteLine(string line)
 
             char noteType = parts[0][0];
 
-            if (noteType == '-')
+            if (UltraStarNote.IsRestType(noteType))
```
Supersonicboss1, line 225: Not super familiar with UltraStar charts, so correct me if i'm wrong, but do you need to reset `_pendingPitchSlide` here, if the note is a rest, to prevent the next phrase after the rest from incorrectly consuming it?

YARG.Core/Chart/Loaders/UltraStar/UltraStarLoader.cs
```diff
@@ -190,66 +251,124 @@ private void ParseNoteLine(string line)
...
+        private void MarkPreviousNoteMelismaJoin()
+        {
+            var partNotes = GetOrCreatePart(_currentPart);
+            for (int i = partNotes.Count - 1; i >= 0; i--)
+            {
+                if (!partNotes[i].IsRest)
+                {
+                    partNotes[i].MelismaJoin = true;
+                    return;
+                }
```
Supersonicboss1, line 312: Again, not super familiar with UltraStar, but if the previous note is a rest, should this return early and not mark any note as `MelismaJoin`?

YARG.Core/Song/Cache/FileCollection.cs
```diff
@@ -53,6 +53,24 @@ public bool FindFile(string name, out FileInfo file)
...
+            foreach (var entry in _entries)
+            {
+                // Keys are already lowercased by the constructor.
+                if (entry.Key.EndsWith(extension, StringComparison.Ordinal) && entry.Value is FileInfo candidate)
+                {
+                    files.Add(candidate);
+                }
+            }
+            return files;
+        }
```
Supersonicboss1, line 72: Not sure about this. I understand that the way UltraStar songs are packaged kind of sucks (no default filename, just a .txt) but i'm wary about doing things like this because they are probably not performant, and this allocates an extra list just to search for files. I think just skipping common readme, licence, or other non-notes filenames, then just blindly matching the first .txt that isn't one of those would probably resolve the majority of cases without requiring allocations.

YARG.Core/Chart/Loaders/UltraStar/UltraStarLoader.cs
```diff
@@ -267,14 +386,24 @@ public SyncTrack LoadSyncTrack()
...
+            var tempos = new List<TempoChange> { new(_bpm / 2.0, -gapSeconds, 0u) };
+
+            // Relative to beat 0, matching the initial entry above.
+            uint tickAtBeatZero = BeatToTick(0);
+            foreach (var (beat, bpm) in _tempoChanges)
+            {
+                uint tick = BeatToTick(beat) - tickAtBeatZero;
+                double time = BeatToTime(beat) - gapSeconds;
+                tempos.Add(new TempoChange(bpm / 2.0, time, tick));
+            }
```
Supersonicboss1, line 402: Notes are shifted based on the GAP parameter in the notes.txt, so does this mean that tempo changes will be out of sync with the notes?

Supersonicboss1 (changes requested):
Finally got around to giving this a review. Should note I have not done performance testing yet, just checked that it did in fact, load the song you DMed me.

YARG.Core/Song/Cache/FileCollection.cs
taynaron: The main issue with that approach is that it's common to have multiple valid song files in a single folder - one for single player, another for multi-track, same as with melody only vs melody + harmony options in YARG. It's messier because the format is messier, but it's ideal to support. I'm open to other ideas though, like "only txt files that start with the song title or 'song' are considered."

YARG.Core/Song/Entries/Ini/SongEntry.IniBase.cs
taynaron: That's fair. I was doing that to avoid changing FileCollection's constructor keys, but now I've added NFC-normalization, which allows subfiles to be properly catalogued. This is a bigger change in that it could potentially affect YARG-style songs, but it works fine on my rather large collection, and should be strictly better in any event.

YARG.Core/Song/Cache/CacheHandler.cs
taynaron: I'll respond in the other comment.

YARG.Core/Chart/Loaders/UltraStar/UltraStarLoader.cs
taynaron: Good catch! I should have seen that. Fixed, test added.

YARG.Core/Chart/Loaders/UltraStar/UltraStarLoader.cs
taynaron: Thanks for that too! Now it only looks at the immediately preceding entry and does nothing if that's a rest. Added a regression test.

YARG.Core/Song/Cache/FileCollection.cs
taynaron: I added a small denylist (readme/license/licence/copying/changelog) checked before opening the file, so the common case — one chart plus one readme — costs zero I/O. Anything not on the list still goes through the content check, so multi-chart folders and oddly-named files stay correct. Let me know if you want to talk more about tradeoffs here. I don't know that we can do much better without more false negatives.

YARG.Core/Chart/Loaders/UltraStar/UltraStarLoader.cs
taynaron: Good eye. I'm trying to match US format spec here, but don't actually have any song that uses this, and couldn't find one. I believe I've fixed it, but am just as happy to remove the bpm change entirely if you like since it's pretty uncommon and at least half of ultrastar-specific players don't even support it.

taynaron:
I did some performance testing, and I'd say that things load about 1.5x as long (or 50% longer than) FoF-style songs, but I don't know how to speed that up w/o missing some songs.

PR #454, opened 2026-09-05
Fix folder sorting for CON multisong packs

asimard1:
## Goal

CON multisong packs are currently not categorized correctly when the library is sorted by folders. By default, they appear under "Unknown Playlist" unless they are placed in a subfolder or have a `pack_name` entry in their DTA. The `pack_name` entry is currently read as `Playlist`, which doesn't seem fitting. More importantly, `Playlist` shouldn't be used to determine the folder name when sorting the library by folders.

Using `pack_name` this way also seems to contradict the goal of folder sorting, which should let users decide how their songs are organized and what their folders are named.

## The fix

The changes are as follows:

1. The `pack_name` entry from DTA files is renamed from `Playlist` to `PackName`. This is currently unused, but if it is used in the future, naming it `PackName` should make its purpose more transparent.
2. When a CON multisong pack is not inside a subfolder, its default folder name is now the name of the CON file itself. This allows users to have a library containing only multisong CON packs without needing to manually create subfolders just to get meaningful names when using folder sorting.

## Additional info

This solves the main issues I was originally trying to address with #411 and #422.

It turns out that video backgrounds **are** supported for CON multisong packs. The videos simply need to be placed alongside the CON file and follow the naming convention `$"{dtanode}.{extension}"`. Since this was one of the reasons I originally wanted to extract my CON packs, this is now much less of an issue.

CON multisong packs also being sorted incorrectly when using folder sorting was another reason I found extracting them useful. This is exactly the issue that this PR wishes to solve, so extracting CON packs would be less necessary than I originally thought.

I still think having the option to support MOGG and `png_xbox` files directly is a worthwhile addition, so I will leave #422 open. The other reasons I mentioned in #422 still apply, particularly avoiding unnecessary conversion and allowing the original files to be used directly. I also believe that supporting the maximum amount of different file types will help the game feel more accessible to all users.

As long as this PR gets merged, though, #422 doesn't serve quite as much of a purpose as it originally did. :)

## Minimal changes

Changing the naming convention is not that big of a deal since making subfolders is not the end of the world, but `pack_name`'s having precedence over a user's named folders shouldn't happen. I would say that renaming pack_name in DTAEntry is the least that could be done to help with this.

### DIFF
--- a/YARG.Core/IO/DTA/DTAEntry.cs
+++ b/YARG.Core/IO/DTA/DTAEntry.cs
@@ -22,7 +22,6 @@ namespace YARG.Core.IO
         public TextSpan? CharterKeys;
         public TextSpan? CharterProStrings;
         public string? Source;
-        public TextSpan? Playlist;
         public TextSpan? LoadingPhrase;
         public int? YearAsNumber;
         public int? YearSecondaryAsNumber;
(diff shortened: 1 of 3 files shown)


asimard1:
Obviously this is not the proper way to do this, and it will be better to do this in a less direct way... I will rename the current "Folder" sorting to "Playlist" sorting, and I will make folder work as it should. But I will do so in a new PR.

asimard1:
This PR was replaced by https://github.com/YARC-Official/YARG/pull/1708

PR #455, opened 2026-09-06
Fix and simplify 6f ghost input check

theli-ua:
Fixes https://discord.com/channels/1086048856678084609/1545669995818917898


Supersonicboss1 (approved):
I should note that I do not play 6F, nor do I own the guitar, so my testing is entirely limited to what is testable on a 5F guitar. Nonetheless, this feels far better, and the code looks fine.

PR #457, opened 2026-09-07
six feet visuals

theli-ua:
core companion for https://github.com/YARC-Official/YARG/pull/1665

### DIFF
--- a/YARG.Core/Game/Presets/ColorProfile.SixFretGuitar.cs
+++ b/YARG.Core/Game/Presets/ColorProfile.SixFretGuitar.cs
@@ -2,7 +2,6 @@ using System.Drawing;
 using System.IO;
 using YARG.Core.Chart;
 using YARG.Core.Extensions;
-using YARG.Core.Input;
 using YARG.Core.Utility;
 
 namespace YARG.Core.Game
@@ -32,7 +31,6 @@ namespace YARG.Core.Game
                     (int) SixFretGuitarFret.White1 => WhiteFret,
                     (int) SixFretGuitarFret.White2 => WhiteFret,
                     (int) SixFretGuitarFret.White3 => WhiteFret,
-                    (int) SixFretGuitarFret.Open => WhiteFret,
                     _ => default
                 };
             }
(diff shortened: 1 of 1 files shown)


YARG.Core/Game/Presets/ColorProfile.SixFretGuitar.cs
```diff
@@ -199,6 +192,10 @@ public void Deserialize(BinaryReader reader, int version = 0)
 
                 BlackNoteStarPower = reader.ReadColor();
                 WhiteNoteStarPower = reader.ReadColor();
+
+                // Note: the Open* color fields are intentionally NOT binary-serialized
+                // (replay format compatibility, no version bump). They persist through
+                // the JSON preset files instead.
```
Supersonicboss1, line 198: To be honest, I don't actually know what the binary serialisation is used for (replays?). But we should nonetheless serialise everything.

YARG.Core/Game/Presets/ColorProfile.SixFretGuitar.cs
```diff
@@ -185,9 +180,7 @@ public void Serialize(BinaryWriter writer)
 
             public void Deserialize(BinaryReader reader, int version = 0)
             {
-                BlackFret = reader.ReadColor();
-                WhiteFret = reader.ReadColor();
-
+                Fret = reader.ReadColor();
```
Supersonicboss1, line 183: This replacing black/white fret will cause old color profile loading issues I think. Version should probably be bumped and compat code should be added.

Supersonicboss1 (changes requested):

YARG.Core/Game/Presets/ColorProfile.SixFretGuitar.cs
theli-ua: I don't think it's worth it. It'll simply fall back to default color . If it would be stable I'd agree with you though

Supersonicboss1 (approved):
To be honest, I really thought that it would cause the replay to fail to load, since binary serialisation is somewhat fickle. But it seems fine, and I'm willing to accept the color profile just not loading.

PR #460, opened 2026-09-09
Commit points when rebasing sustains

Supersonicboss1:
This also rebases sustains early when SP ends, so that it can capture the SP state in AddScore before it is actually disabled. This means it rebases twice here, but the second should be skipped.

### DIFF
--- a/YARG.Core.UnitTests/Engine/SustainListTests.cs
+++ b/YARG.Core.UnitTests/Engine/SustainListTests.cs
@@ -114,7 +114,6 @@ public class SustainListTests
         {
             Assert.That(sustain.Note, Is.SameAs(note));
             Assert.That(sustain.BaseTick, Is.EqualTo(note.Tick));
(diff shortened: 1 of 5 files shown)


04meta:
I was going to ask to check if this also applied to bandmates changing their star power status, but it turns out that sustain points don't actually account for bandmate SP to begin with; the most you can get is 8x or 12x for your *own* SP.

I should probably file another issue about that, but I'm not sure whether to do it in YARG or YARG.Core...

wyrdough:
> I was going to ask to check if this also applied to bandmates changing their star power status, but it turns out that sustain points don't actually account for bandmate SP to begin with; the most you can get is 8x or 12x for your _own_ SP.
> 
> I should probably file another issue about that, but I'm not sure whether to do it in YARG or YARG.Core...

Anything you earn as a result of bandmate SP is recorded as band bonus score, it does not affect individual player scores.

PR #461, opened 2026-09-09
Add orthogonal projection to camera preset

theli-ua:


YARG.Core/Game/Presets/CameraPreset.cs
```diff
@@ -28,6 +28,13 @@ public partial class CameraPreset : BasePreset
         [SettingRange(-3f, 3f)]
         public float CurveFactor = 0.5f;
 
+        [SettingType(SettingType.Toggle)]
+        public bool Orthographic = false;
+
+        [SettingType(SettingType.Slider)]
+        [SettingRange(0.1f, 10f)]
+        public float OrthographicSize = 2f;
```
wyrdough, line 36: Just so it's in my brain, we should probably have a means of hiding settings that won't be used because of a toggle being off...

wyrdough (approved):

PR #463, opened 2026-09-11
fix(bot): rebuild guitar bot held-fret mask each tick to drop stale bits

theli-ua:
UpdateBot only rewrote EffectiveButtonMask on a new bot note press. An extended-sustain fret bit from a bot note whose sustain had ended lingered, breaking the hold check of a still-active overlapping sustain whose target fret was lower

Only affects bot code


theli-ua:
https://discord.com/channels/1086048856678084609/1547282584570953798  discord thread has chart for repro attached

